PR #8, opened 2019-02-19
Refactor files

macrozone:
based on https://github.com/unchainedshop/unchained/pull/7, so see and merge that first 

introduces unchained:core-files

which hosts all file collection related stuff

GOAL:

be able to make it easier to customize file collections.

e.g. having a config like this:

```
const fileConfig = {
   order_documents: {
      storage: {
         type: "filesystem",
         path: "/path/to/rootFolder"
      },
        maxSize: 10485760,
   },
   media: {
     storage: {
       type: "gridFS"
     },
     maxSize: 10485760,
     extensionRegex: /png|jpg|jpeg/i,
   },
   avatars: {
      storage: {
        type: "aws-s3",
        credentials: { ...},
        bucket: "my-first-shop-bucket"
      },
       maxSize: 10485760,
     extensionRegex: /png|jpg|jpeg/i,
    }


}
```


pozylon:
I really like the approach, cleaner dependencies and better separation, do you already have an idea on how you would pass the config to the core-files package? and would it make sense to configure a file provider per collection?

macrozone:
@pozylon asked myself the same.

maybe you can pass the whole config into `startPlatform` https://github.com/unchainedshop/unchained/blob/master/examples/minimal/boot.js#L66

becasue you already receive some config there. 

Problem is, that every collection is already initialized at that point.

Or switch to some config singleton, but still then it depends on when you configure it, wheter it will have an effect or no...  so not very nice.

It would best, if everything would be initialized on startPlatform, including collections, but that makes imports annyoing... i will think about it

pozylon:
LGTM

PR #12, opened 2019-02-24
feat: gridfs for file-collections and new settings-api

macrozone:
this pr enables to configure gridfs, instead of local file system for files. local file system is still the default, but i would suggest to make gridfs the default as it is much more convenient for smaller projects as you don't have to mess around with persistent volumes.

I also have a suggestion about a new settings-api using Meteor.settings

i decided to use Meteor.settings because it is available across packages when they load. Without Meteor.settings, its rather tricky to have a configuration in place before the app code initializes. 

To make it more agnostic and allow to get config from other places, e.g. a database or ENV_VARS, i wrapped it in a function.

### Settings API

E.g. given these settings:

```
// meteor-settings.json
{
  "unchained": {
    "files": {
      "media": {
         "maxSize": 10485760
      }
    }
  }
}
```

you can now access `maxSize` with this:

```
import { getSetting } from 'meteor/unchained:core-settings';

const maxSize = getSetting('files.media.maxSize')

// it uses lodash.get, so you can pass a path as an array
const maxSize = getSetting(["files", "media", "maxSize"])

// and you can specify a default value

const maxSize = getSetting("files.avatars.maxSize", 10485760);

```

### configure FileCollections

to set gridfs for all collections, use this:

```
{
  "unchained": {
    "files": {
      "default": {
        "storage": {
          "type": "gridfs"
        }
      }
    }
  }
}
```

all collections will use `default`, unless specified directly:

```
{
  "unchained": {
    "files": {
      "media": {
        "maxSize": 10485760
      }
    }
  }
}
```


pozylon:
Thank you very much for this PR, i will review the code in the next few days. I've found 2 things that you might change:

1. If i set "unchained.files.media.maxSize" it should be the highest priority and override all default values, that makes absolutely sense. But let's assume we use createFileCollection and pass maxSize as a "customSetting": createFileCollection("avatar", { "maxSize": 1024 }), then I set unchained.files.default.maxSize to 10485760 in the userSettings. In that case, I would prefer having only overwritten the maxSize for all file collections that don't have a customSetting for maxSize, e.G:

```
const userSettings = , );
	  const mergedSettings = {
	    ...createDefaultSettings(),
            ...getSetting('files.default'),
            ...customSettings,
	    ...getSetting(['files', collectionName]),
	  };
```

2. "i decided to use Meteor.settings because it is available across packages when they load. Without Meteor.settings, its rather tricky to have a configuration in place before the app code initializes.". I think through the "getSetting" function you just introduced a solution to that app code initialization issue. getSetting could load the settings if that has not been done yet, it could even use dotenv or "fs" to load configs and we could have some sort of ENV-only configuration approach:

unchained.files.media.maxSize -> UNCHAINED.FILES.MEDIA.MAX_SIZE

pozylon:
Of cours the env only approach would be opt-in, having a json for configuration is more handy to many people.

macrozone:
> Thank you very much for this PR, i will review the code in the next few days. I've found 2 things that you might change:
> 
> 1. If i set "unchained.files.media.maxSize" it should be the highest priority and override all default values, that makes absolutely sense. But let's assume we use createFileCollection and pass maxSize as a "customSetting": createFileCollection("avatar", { "maxSize": 1024 }), then I set unchained.files.default.maxSize to 10485760 in the userSettings. In that case, I would prefer having only overwritten the maxSize for all file collections that don't have a customSetting for maxSize, e.G:
> 
> ```
> const userSettings = , );
> 	  const mergedSettings = {
> 	    ...createDefaultSettings(),
>             ...getSetting('files.default'),
>             ...customSettings,
> 	    ...getSetting(['files', collectionName]),
> 	  };
> ```
> 1. "i decided to use Meteor.settings because it is available across packages when they load. Without Meteor.settings, its rather tricky to have a configuration in place before the app code initializes.". I think through the "getSetting" function you just introduced a solution to that app code initialization issue. getSetting could load the settings if that has not been done yet, it could even use dotenv or "fs" to load configs and we could have some sort of ENV-only configuration approach:
> 
> unchained.files.media.maxSize -> UNCHAINED.FILES.MEDIA.MAX_SIZE

macrozone:
> Thank you very much for this PR, i will review the code in the next few days. I've found 2 things that you might change:
> 
> 1. If i set "unchained.files.media.maxSize" it should be the highest priority and override all default values, that makes absolutely sense. But let's assume we use createFileCollection and pass maxSize as a "customSetting": createFileCollection("avatar", { "maxSize": 1024 }), then I set unchained.files.default.maxSize to 10485760 in the userSettings. In that case, I would prefer having only overwritten the maxSize for all file collections that don't have a customSetting for maxSize, e.G:
> 
> ```
> const userSettings = , );
> 	  const mergedSettings = {
> 	    ...createDefaultSettings(),
>             ...getSetting('files.default'),
>             ...customSettings,
> 	    ...getSetting(['files', collectionName]),
> 	  };
> ```
> 1. "i decided to use Meteor.settings because it is available across packages when they load. Without Meteor.settings, its rather tricky to have a configuration in place before the app code initializes.". I think through the "getSetting" function you just introduced a solution to that app code initialization issue. getSetting could load the settings if that has not been done yet, it could even use dotenv or "fs" to load configs and we could have some sort of ENV-only configuration approach:
> 
> unchained.files.media.maxSize -> UNCHAINED.FILES.MEDIA.MAX_SIZE

ok, will do that with `..getSetting(['files', collectionName]),`

the env only approach can be done in a follow up approach.

macrozone:
@pozylon i merge now the settings, can you check?

packages/core-settings/package.js
```diff
@@ -0,0 +1,28 @@
+Package.describe({
+  name: 'unchained:core-settings',
+  version: '0.0.1',
```
pozylon: Please use same version as other packages, in the beginning we just publish all packages together with search & replace

packages/core-settings/package.js
```diff
@@ -0,0 +1,28 @@
...
+  // Brief, one-line summary of the package.
+  summary: '',
+  // URL to the Git repository containing the source code for this package.
+  git: '',
+  // By default, Meteor will default to using README.md for documentation.
+  // To avoid submitting documentation, set this field to null.
+  documentation: 'README.md',
```
pozylon: Please see other package.js: A summary and a non-empty readme.md is required to publish it on atmosphere

macrozone:
@pozylon ok changes done

pozylon:
LGTM

PR #14, opened 2019-02-25
Quotations alpha

pozylon:
This does not only add an alpha version of quotations but it also has a breaking change for core-documents, see
https://github.com/unchainedshop/unchained/commit/4bbb8f09e5d4c3043ff573a6145c50faee24628a


pozylon:
@macrozone ping

PR #28, opened 2019-03-04
Basic Quotations Implementation

pozylon:
New logic when adding products to a cart:

- addCartProduct of a ConfigurableProduct now resolves to the concrete product if enough variation vector parameters provided through product configuration parameters. The source productId is saved into context.origin, the variation configuration is stored on the item along other product configuration parameters

- addCartQuotation resolved to the actual product and adds that to the cart. It uses a transform method of the quotation plugin system to transform a quotationConfiguration to a productConfiguration. The source quotationId is saved into context.origin

It's also possible to chain:

- addCartQuotation is called with quotation Y
-> resolves to a Matrix product X with a specific configuration
-> specific configuration is handed to vector logic trying to find a distinct concrete product Z
-> bundle product Z is resolved

Now the cart looks like this:
- 1 x Bundle Z


pozylon:
Breaking:

- addCartProduct of a ConfigurableProduct now resolves to the concrete product if enough variation vector parameters provided through product configuration parameters. The originating productId is saved into context.origin, the variation configuration is stored on the item along other product configuration parameters
- addCartQuotation resolved to the actual product and adds that to the cart. It uses a transform method of the quotation plugin system to transform a quotationConfiguration to a productConfiguration. The originating quotationId is saved into context.origin

pozylon:
Missing step:

- addCartProduct of a BundleProduct should behave different and not explode on addition to cart. But if the bundle wants to get "individualized" it needs to get "exploded" into it's parts first.

pozylon:
It's also possible to chain:

- addCartQuotation is called with quotation Y
-> resolves to a Matrix product X with a specific configuration
-> specific configuration is handed to vector logic trying to find a distinct concrete product Z
-> bundle product Z is resolved

Now the cart looks like this:
- 1 x Bundle Z

If somebody calls explode on Bundle Z, the explode transform will run through together with the special configuration, manipulating the cart to:

- 5 x Simple Product A
- 10 x Simple Product B
- 1 x Simple Product C

PR #39, opened 2019-03-20
feat: new properties for send-mail delivery provider

macrozone:


packages/core-delivery/plugins/send-mail.js
```diff
@@ -25,11 +25,13 @@ class SendMail extends DeliveryAdapter {
     return type === 'SHIPPING';
   }
 
-  isActive() { // eslint-disable-line
+  isActive() {
+    // eslint-disable-line
```
pozylon: This will lead to more warnings I guess :)

macrozone: not sure why the reformat ;-)

pozylon: prettier but shittier 🤟

pozylon:
What if we just forward the whole order object to the template? It would be more generic and totally delegate what data to show and how to format to the actual template.

pozylon:
```
return director.sendMessage({
      template: 'shop.unchained.send-mail',
      attachments,
      meta: {
        mailPrefix: `${order.orderNumber}_`,
        from: this.getFromAddress(),
        to: this.getToAddress(),
        cc: this.getCCAddress(),
        transactionContext,
        order
      }
    });
```

macrozone:
we should also rename meta in all emails, because meta is ambiguos.

i would just call it `props`

forwarding order does not work so good, because the template language used can't resolve the helpers.  I would also suggest to swtich the template language as mustache is too limited in my opinion. Better use handlebars or so. Maybe we can even use blaze or handlebars in a way, so that object helpers are resolved, which would be very nice

pozylon:
Ok so maybe more like this without props and meta:

```
return director.sendMessage({
      template: 'shop.unchained.send-mail',
      attachments,
      mailPrefix: `${order.orderNumber}_`,
      from: this.getFromAddress(),
      to: this.getToAddress(),
      cc: this.getCCAddress(),
    }, { transactionContext,
        order
});
```

Yeah, if handlebars can do that, it would match much better.

pozylon:
Going to merge this, let's do the "Refactor E-Mail" in another PR

Issue #45, opened 2019-03-21
type error: updateProduct allows to change type, yet you have to use db type instead of graphql type


macrozone:
calling mutation `updateProduct(productId, product)`

with 

```
product = {
   type: "SimpleProduct"
}

```

will write `SimpleProduct` directly into the db, whereas when creating a product with `createProduct`, it will write `SIMPLE_PRODUCT` in db

PR #46, opened 2019-03-21
fix https://github.com/unchainedshop/unchained/issues/45

macrozone:
fixes https://github.com/unchainedshop/unchained/issues/45


pozylon:
Good one!

PR #47, opened 2019-03-21
feat: add meta type for products

macrozone:


pozylon:
That makes sense given that meta is part of the simple schema but shouldn't we introduce meta to the input types and create/update mutations as well?

macrozone:
> That makes sense given that meta is part of the simple schema but shouldn't we introduce meta to the input types and create/update mutations as well?

probably. I updated them server side directly in the project, so i did not need the mutation

macrozone:
can we merge this?

PR #48, opened 2019-03-21
relax orderitem

macrozone:
if you remove products, but have orders with these products around, you can't fetch the order or cart anymore, it will throw a server error.

it's better to relax that here so that clients can react to this situation.

### DIFF
--- a/packages/api/schema/types/order/item.js
+++ b/packages/api/schema/types/order/item.js
@@ -19,10 +19,10 @@ export default [
 
     type OrderItem {
       _id: ID!
-      product: Product!
+      product: Product
       order: Order!
       quantity: Int!
-      originalProduct: Product!
+      originalProduct: Product
       quotation: Quotation
       unitPrice: Money
       total(category: OrderItemPriceCategory): Money


pozylon:
I would like to tackle this differently, as products are extremely central they should not get deleted from the database (technically) at all. For a user it should always be possible to see the order history and the products she/he once ordered years ago.

I have a very long living task on my site named: "Introduce deletion flags". The idea is that we get rid of all Collection.remove calls for order-history-relevant master data. But maybe there is other concepts?

macrozone:
actually thats a valid point. But i am not sure if its a good idea to enforce that on the graphql surface. Clients that encounter a cart where a product has vanished somehow are now in a broken state. they just can't fetch that cart anymore.

you can't solve this integrity problem anyway properly without a db that enforces integrity. ;-)

Anyway, its basically a problem that i have now during development. i just have to remove all orders as well if i delete products.

pozylon:
We could introduce product deletion flags and at the same time write a migration that fixes the productId references with a dummy product with state deleted that says: "Deleted Product" or something ^^

It's not perfect, though ^^

Issue #58, opened 2019-04-30
add lastContact to user


macrozone:
a user has lastBillingAddress and lastDeliveryAddress, but not lastContact.

I will send in a pull request that saves lastContact and exposes it as resolver.

also I noticed that the naming is inconsistent.

`Contact` as in `lastContact` has `telNumber` and `emailAddress`, while in other places, `phoneMobile` is used for phone

Maybe the whole contact thing has to be removed and email and phone should be added to deliveryAddress (or billingAddress).

PR #59, opened 2019-04-30
feat: expose lastContact on user

macrozone:
fixes https://github.com/unchainedshop/unchained/issues/58


pozylon:
Thank you very much, LGTM.

I just refactored Users.updateLastContact a bit so we only update the user once and the updatedAt gets set correctly

PR #60, opened 2019-04-30
Consistent upsert text

pozylon:
Various small refactorings in controlpanel and internal upsertLocalizedText  methods


pozylon:
I'm actually not very happy with this PR :S

PR #62, opened 2019-05-02
relax ID restriction

macrozone:
this PR removes the validation of `SimpleSchema.RegEx.Id` on all Ids and replaces it with simple String.

## Reasoning

While it is generally preferable to have non-guessable IDs, it sometimes make sense to have predefined IDs, in particular if you import products or similar from external sources.

This change does not change any default behaviour, it just allows to pass any string s ID when you create a Product


pozylon:
makes sense and LGTM

PR #65, opened 2019-05-07
Refactor Master Data Entity API

pozylon:
- Introduces missing convenience helpers to create master data entities
- Handles authorId's consistent across all master data entities
- Adds authorId to various entities in order to backtrack master data changes to users
- Moves all migrations into their own migrations folder separating it from the schema


pozylon:
I'm going to wait with merging this until we have basic integration tests

PR #66, opened 2019-05-07
Controlpanel exportability

pozylon:
Minor changes that allow to next export the controlpanel. This helps support cases embedding the controlpanel in the engine backend project


pozylon:
@macrozone I'm planning to embed the controlpanel in the vps project's engine

PR #67, opened 2019-05-07
Integration Tests

pozylon:
- [x] Find a way to test the GraphQL API's by introspecting an In-Memory MongoDB that works seamless together with a meteor project
- [x] Find out how --watch mode can be used in a way that the meteor server and mongod don't restart
- [x] Add Basic Integration Tests for Shop Setup and Authentication


pozylon:
This PR is getting too big, Let's merge the test setup and the first few tests so everybody can start profit

tests/helpers.js
```diff
@@ -0,0 +1,58 @@
...
+) {
+  try {
+    const { insertedId } = await this.insertOne(doc, ...args);
+    return this.findOne({ _id: insertedId }, ...args);
+  } catch (e) {
+    return this.findOne({ _id: doc._id }, ...args);
+  }
+};
+
+module.exports.setupDatabase = async () => {
```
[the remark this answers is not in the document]
pozylon: I think I can't unless I run the code through babel but maybe I'm wrong, will try

tests/auth-anonymous.test.js
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  afterAll(async () => {
+    await connection.close();
+  });
+
+  it('login with password and logout', async () => {
+    const { data: { loginWithPassword } = {} } = await apolloFetch({
+      query: /* GraphQL */ `
+        mutation {
```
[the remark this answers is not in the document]
pozylon: generally a good idea, especially when used together with an example that implements a very basic storefront, the actual example would then run on only tested code, but I'm not sure if or how to share these files with customer projects?

pozylon:
Thanks for the review!

tests/helpers.js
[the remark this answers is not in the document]
schmidsi: You probably need a Jest plugin or something. Maybe this helps: https://jestjs.io/docs/en/getting-started.html#using-babel

tests/auth-anonymous.test.js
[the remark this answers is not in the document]
schmidsi: Put them into [`packages/api/queries`](https://github.com/unchainedshop/unchained/tree/master/packages/api)?

package.json
```diff
@@ -18,17 +18,23 @@
...
   "scripts": {
     "lint": "eslint packages --cache --fix --ext=jsx --ext=js",
+    "test": "jest",
+    "test:watch": "DEBUG=jest-mongodb:* jest --watch --verbose",
     "bump": "./bump.sh",
     "dev": "./dev.sh"
   },
   "devDependencies": {
+    "@shelf/jest-mongodb": "^1.0.1",
+    "apollo-fetch": "^0.7.0",
```
macrozone: apollo-fetch is deprecated as it seems. Maybe this would also work? https://github.com/prisma/graphql-request

tests/auth-anonymous.test.js
```diff
@@ -0,0 +1,91 @@
...
+            id
+            token
+          }
+        }
+      `
+    });
+    expect(loginWithPassword).toMatchObject({
+      id: 'admin'
+    });
+  });
```
macrozone: maybe group all login tests further:

```
describe('authentication', () => {
    describe("mutation.loginWithPassword", () => {
        it("allows to login with given username and password", () => {
           // your test
        })

        it("throws when username or password is wrong", () => {

        })

          it("returns a user object", () => {
         
          })

          // etc.

    })
})

File: tests/helpers.js [hunk 5]
```diff
@@ -0,0 +1,58 @@
...
+Collection.prototype.findOrInsertOne = async function findOrInsertOne(
+  doc,
+  ...args
+) {
+  try {
+    const { insertedId } = await this.insertOne(doc, ...args);
+    return this.findOne({ _id: insertedId }, ...args);
+  } catch (e) {
+    return this.findOne({ _id: doc._id }, ...args);
+  }
```
macrozone: it is generally advisable to not do too much magic, maybe its better to be precice in tests on insert or find

package.json
pozylon: Oh wow I didn't see that it is deprecated, but in that case I'd favor just using Apollo Link because we use the Apollo libraries on the server, too, I would create a convenience fetch method with the same signature as apolloFetch and then invoke it that way (apollo docs):

```
import { execute, makePromise } from 'apollo-link';
import { HttpLink } from 'apollo-link-http';
import gql from 'graphql-tag';

const uri = 'http://api.githunt.com/graphql';
const link = new HttpLink({ uri });

const operation = {
  query: gql`query { hello }`,
  variables: {} //optional
  operationName: {} //optional
  context: {} //optional
  extensions: {} //optional
};

// execute returns an Observable so it can be subscribed to
execute(link, operation).subscribe({
  next: data => console.log(`received data: ${JSON.stringify(data, null, 2)}`),
  error: error => console.log(`received error ${error}`),
  complete: () => console.log(`complete`),
})

// For single execution operations, a Promise can be used
makePromise(execute(link, operation))
  .then(data => console.log(`received data ${JSON.stringify(data, null, 2)}`))
  .catch(error => console.log(`received error ${error}`))
```

tests/auth-anonymous.test.js
pozylon: true

tests/helpers.js
pozylon: I had race conditions because the top level test suits all tried to create the admin user at the same time, i fixed it by turning things around:

before:
await findOne -> conditionally await insert

now:
try await insert -> await findOne even on catch

You're definitively right, this is a little bit of magic. Maybe I tried to solve things the wrong way, how could I prepare the db for the tests in a more safe way?

pozylon: There is --maxConcurrency=

PR #68, opened 2019-05-07
fix: if an order contains an item with price 0 and discount is applied, the total price becomes NaN

macrozone:
reproduce:

add item with price 0 to cart,
add item with non-0 price
add discount on cart

total of cart will be NaN (or 0)

please verify the implementation

### DIFF
--- a/packages/core-pricing/plugins/order-discount.js
+++ b/packages/core-pricing/plugins/order-discount.js
@@ -30,13 +30,6 @@ class OrderItems extends OrderPricingAdapter {
       const pricing = item.pricing();
       const tax = pricing.taxSum();
       const gross = pricing.gross();
-      if (totalItemsAmount === 0 || gross - tax === 0) {
-        return {
-          ratio: 0,
-          taxDivisor: 1
-        };
-      }
-
       return {
         ratio: gross / totalItemsAmount,
         taxDivisor: gross / (gross - tax)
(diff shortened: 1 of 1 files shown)


packages/core-pricing/plugins/order-discount.js
```diff
@@ -30,6 +30,13 @@ class OrderItems extends OrderPricingAdapter {
       const pricing = item.pricing();
       const tax = pricing.taxSum();
       const gross = pricing.gross();
+      if (totalItemsAmount === 0 || gross - tax === 0) {
+        return {
```
pozylon: The implementation is correct but if you return 0 & 1 here the result of this code here:
```javascript
shares.forEach(({ ratio, taxDivisor }) => {
        const shareAmount = discountAmountToSplit * ratio;
        const shareTaxAmount = shareAmount - shareAmount / taxDivisor;
        discountAmount += shareAmount;
        taxAmount += shareTaxAmount;
      });
```

would effectively translate to this:
```javascript
shares.forEach(({ 0, 1 }) => {
        const shareAmount = discountAmountToSplit * 0; // -> shareAmount = 0
        const shareTaxAmount = shareAmount - 0 / 1; // ->shareTaxAmount = 0
        discountAmount += shareAmount; // -> discountAmount stays the same
        taxAmount += shareTaxAmount; // -> taxAmount stays the same
      });
```

So it's safe to .filter out the share if one of the two variables evaluate to NaN and it would result in less lines of code but it's up to you to change this, for me it's fine and you can merge at will.

pozylon:
Thanks for the PR

packages/core-pricing/plugins/order-discount.js
pozylon: @macrozone ping

Issue #26, opened 2019-02-27
BUG: can't add multipe products to cart in one request


macrozone:

pozylon:
maybe related: https://github.com/apollographql/apollo-link-state/issues/40

macrozone:
@pozylon i tried that, but it resulted in inconsitent results, i sometimes got only one product and in some occasions three products, but i sent two. Have to revisist that, something was wrong.

macrozone:
the bug is that `addCartProduct` will create a new order if no orderId is given or it tries to fetch one from the user. If you fire multiple `addCartProduct` at the same time, it can happen that both will create a new Order.

solution would be to adress the cart never by its orderId, but with something that is always consistent. Consider this:

- if you only allow 1 cart, its straight forward. The cart is always `Orders.findOne({userId, isCart: true})` (i know `isCart` does not exist, its in fact `{countryCode, status: { $eq: OrderStatus.OPEN }}` but you get the idea)

- if you want to have multiple carts, you have to give everyone a certain identifier, like its index (0 is always the default cart). If a user wants to add an item to a certain cart, he has to address this cart anyway somehow

Another solution would be to use mongodb transaction and try to lock the db from start of the mutation until the end. (or lock somehow the mutation. Meteor's older methods did indeed something like that: they were always sequential for a given user)

PR #76, opened 2019-06-21
feat: allow to add multiple cart products in one request

macrozone:
workaround for https://github.com/unchainedshop/unchained/issues/26

adds a new new mutation `addMultipleCartProducts`

does not fix the underlying concurrency problem though, but i think this is one obvious usecase were you can run into that problem.


pozylon:
it feels like we would add a mutation for a very specific edge case just because apollo can't handle mutations serially the way it should be, but there are no alternatives at the moment...

The graphql specification explicitly states that mutations have side effects.

But is it really a core API task to allow adding multiple products in one call, alternatively this additional mutation could just sit in the "extensions" part for a specific unchained-based project like the one you're currently implementing.

pozylon:
I think it's okay to have the mutation enlarging the API surface as adding stuff to the cart is very central to the engine.

PR #82, opened 2019-07-23
Paths

pozylon:
This PR allows to query "directed paths" https://en.wikipedia.org/wiki/Path_(graph_theory) for the purpose of breadcrumb rendering

- [x] Assortment.assortmentPaths
- [x] Product.assortmentPaths


packages/core-assortments/db/helpers/build-paths.js
```diff
@@ -0,0 +1,56 @@
...
+  const walk = async (assortmentId, initialPaths = [], childAssortmentId) => {
+    const assortmentLink = await resolveAssortmentLink(
+      assortmentId,
+      childAssortmentId
+    );
+    if (!assortmentLink) return initialPaths;
+
+    const subAsssortmentLinks = await Promise.all(
+      assortmentLink.parentIds.map(async parentAssortmentId => {
+        return walk(parentAssortmentId, initialPaths, assortmentId);
```
macrozone: https://www.youtube.com/watch?v=tbNlMtqrYS0

packages/core-assortments/db/helpers/build-paths.js
```diff
@@ -0,0 +1,56 @@
...
+      })
+    );
+
+    if (subAsssortmentLinks.length > 0) {
+      return subAsssortmentLinks.map(subAsssortmentLink => {
+        return [...subAsssortmentLink.flat(), assortmentLink, ...initialPaths];
+      });
```
macrozone: you could do the mapping directly in subAsssortmentLinks, since you map twice over same-sized arrays

pozylon: I've been looking at that code since 10 minutes and i can't follow 🤪

PR #83, opened 2019-07-30
WIP: Native async/await

pozylon:
We currently use meteor/promise but as async and await are native since meteor 1.7, the explicit usage is not needed anymore. This PR will removes strict dependency to meteor/promise (usually Promise.await calls)


pozylon:
Became impossible to merge :)

PR #113, opened 2019-10-22
Datatrans Plugin

pozylon:


schmidsi:
Maybe you could also remove the OrderPaymentCrypto? For my Coinbase implementation I also abused the Paypal base class, since there is a challenge-response thing going on. But going forward on this thinking, what abstraction do these types give?

My suggestion would be something like: They define a kind of "workflow". Where invoice do not delegate to an external provider, generic does.

(Sorry, this is not well thought through 😊 )

pozylon:
Yes somehow, I found out that Postfinance, Paypal, Stripe, Braintree & Datatrans etc. all work the same: They usually work with webhooks and without in redirect and embedded modes and the payment is done kind of before the actual checkout.

For invoice kind of payment the checkout triggers the next steps (one could assume invoice prepaid is different but it's not as you have to block the order and send an e-mail with payment tutorial)

For the crypto thing I though of doing crypto payments without coinbase (coinbase would be type generic) by generating a receiving wallet address upfront where people can deposit.

schmidsi:
> For the crypto thing I though of doing crypto payments without coinbase (coinbase would be type generic) by generating a receiving wallet address upfront where people can deposit

True, Coinbase works the same as the other generics. For crypto, one should probably first actually implement such a functionality. In the real world, there will be probably different approaches to receive cryptos without intermediaries: Every chain has its own rules, there can be receiving smart-contracts, escrow contracts, self-hosted microservices (with higher security standards), etc. I'd remove the `OrderPaymentCrypto` for now and just leave a comment until we have a reference implementation.

pozylon:
@schmidsi Crypto has been removed

PR #119, opened 2019-11-13
Extend Filter plugin system

pozylon:
This will extend the filter plugin system in a way that allows for search backends to take over various parts of the faceting filter.

Major:

- Adds a new filter plugin called "shop.unchained.filters.local-search" that uses ProductText fields as a full text search index with mongodb native $search

Breaking Changes:

- transformSelector becomes async transformProductSelector
- transformProductSelector is now called once without a key where it is used to dynamically adjust the preselected filters for a query result, return lastSelector if !key if you don't need that funcion


pozylon:
```
query search($filterQuery: [FilterQueryInput!], $queryString: String) {
  search(
    filterQuery: $filterQuery
    queryString: $queryString
    includeInactive: false
  ) {
    totalProducts
    filteredProducts
    products {
      _id
      texts {
        title
        subtitle
        brand
        vendor
      }
    }
    filters {
      examinedProducts
      filteredProducts
      isSelected
      definition {
        key
      }
      options {
        filteredProducts
        isSelected
        definition {
          value
          texts {
            title
          }
        }
      }
    }
  }
}

```
{
  "queryString": "helm",
  "filterQuery": [{"key": "__CATEGORIES__", "value": "K4"}]
}
```

```

Issue #118, opened 2019-11-13
[core] Log/Record IP, remote port and user_agent_string for crucial actions


schmidsi:
Crucial actions:

(Everywhere where `... logFields` is used)

- [ ] order-deliveries
- [ ] order-payments
- [ ] orders
- [ ] quotations

pozylon:
That's cool

schmidsi:
@pozylon Touching this subject again and also in connection to #316. Why do we have `logFields` on some models and [on others, the logs are fetched from the `logs` collection](https://github.com/unchainedshop/unchained/blob/master/packages/core-users/db/helpers.js#L139)?

I'd like to streamline this. We might put all the logging functionality into the events. The question is, who reads data from the logs?

pozylon:
I think logs should just be like timestamp + level + message + meta and could be anything, it's just what you can see on stdout and you can deactivate when you don't need to see the logs in the controlpanel.

- Events should be predefined like timestamp + EVENT_NAME + payload without arbitrary messages. If the event is PRODUCT_ADDED_TO_CART, {userId, productId, quantity.. } it can be translated by a UI like "User X added Product Y with quantity 5 to the Order Z"...
- The information in logFields is something you want to store for audit purposes like when the police knocks at your door and asks what the IP was of that user that has been paying by credit card on time/date...

So depending on how transient we will make the events, logFields could be replaced by events. An order could trigger these kind of events for ex.
ORDER_FAILED_PAYMENT (ip, user, reason...),,
ORDER_FAILED_PAYMENT (ip, user, reason...),,
ORDER_SUCCESSFUL_PAYMENT (ip, user...),

schmidsi:
Thanks for the explanation. One thing that we need to keep in mind is the ability to access logs from the code:

For example: The order confirmation mail to the support team needs to include the IP address of the buyer. How can we access the logs and find this IP?

My suggestion is to provide an interface from the logging module. Something like:

```js

LogModule.findLogs({ type: "ADD_TO_CART_MUTATION", cartId: "asdf" });

```

pozylon:
Mutation in the type name is propably wrong because it should be independent from the api package.

I'd strictly distinguish:
Logs -> Arbitrary messages, Untyped with Severity
Events -> Typed, Semantic, no Severity/Level

So if you need that I'd go with
```
EventModule.findEvents({ type: "PRODUCT_ADDED_TO_CART", cartId: "asdf" });
```

[...]

PR #134, opened 2020-01-14
Scribble of posible tracking

schmidsi:
A rough (not working yet) scribble how to solve #118. Thoughts so far?

### DIFF
--- a/.graphqlconfig
+++ b/.graphqlconfig
@@ -1,0 +1,11 @@
+{
+  "projects": {
+    "unchained": {
+      "extensions": {
+        "endpoints": {
+          "dev": "http://localhost:4010/graphql"
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


pozylon:
Okay, cool!

Enhancing the apollo server context with remoteAddress, port and userAgent -> Absolutely agreed, that is data that should be available for any resolvers

Storing the tracking logs inside the collection's logs field -> Forwarding the tracking data through the system so we can store it in the logs fields doesn't seem right to me.


Actions that origin from a resolver are currently already logged with context data (userId, ...), like:
```
2020-01-14T16:57:49.395Z [unchained] info: query products: 10 8 includeDrafts  { userId: 'KymBd5QvBWpKuRJGb' }
```

Also those logs are currently stored in the db and there are already queries like User.logs and Order.logs that allow fetching the logs that belong to a certain user or order. What if we log something like this for all API requests and call it "access log" similar to web server logs?

```
2020-01-14T16:57:49.395Z [unchained:api] info: "1.1.1.1:56124" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_2) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.4 Safari/605.1.15" "USERID..." query products: 10 8 includeDrafts
```

Issue #37, opened 2019-03-05
[core] Add tests


pozylon:
# Summary
- [ ] integration tests for the api package

## Integration Tests

Queries:
- [x] me: User
- [x] users(...): [User!]!
- [x] user(...): User
- [x] products(...): [Product!]!
- [x] product(...): Product
- [x] productCatalogPrices(...): [ProductPrice!]!
- [x] translatedProductTexts(...): [ProductTexts!]!
- [x] translatedProductMediaTexts(...): [ProductMediaTexts!]!
- [x] translatedProductVariationTexts(...): [ProductVariationTexts!]!
- [x] languages(...): [Language]!
- [x] language(...): Language
- [x] countries(...): [Country!]!
- [x] country(...): Country
- [x] currencies(...): [Currency!]!
- [x] currency(...): Currency
- [x] deliveryProviders(...): [DeliveryProvider!]!
- [x] deliveryProvider(...): DeliveryProvider
- [x] deliveryInterfaces(...): [DeliveryInterface!]!
- [x] warehousingProviders(...): [WarehousingProvider!]!
- [x] warehousingProvider(...): WarehousingProvider
- [x] warehousingInterfaces(...): [WarehousingInterface!]!
- [x] paymentProviders(...): [PaymentProvider!]!
- [x] paymentProvider(...): PaymentProvider
- [x] paymentInterfaces(...): [PaymentInterface!]!
- [x] orders(...): [Order!]!
- [x] order(...): Order
- [x] logs(...): [Log!]!
- [x] shopInfo: Shop!
- [x] assortments(...): [Assortment!]!
- [x] assortment(...): Assortment
- [x] translatedAssortmentTexts(...): [AssortmentTexts!]!
- [x] translatedFilterTexts(...): [FilterTexts!]!
- [x] filters(...): [Filter!]!
- [x] filter(...): Filter
- [x] productReviews(...): [ProductReview!]!
- [x] productReview(...): [ProductReview!]!
- [x] quotations(...): [Quotation!]!
- [x] quotation(...): Quotation
- [ ] search: SearchResult

Mutations:
- [x] loginWithPassword(...): LoginMethodResponse
- [x] createUser(...): LoginMethodResponse
- [x] changePassword(...): SuccessResponse
- [x] forgotPassword(...): SuccessResponse
- [x] resetPassword(...): LoginMethodResponse
- [x] logout(...): SuccessResponse
- [x] verifyEmail(...): LoginMethodResponse
[...]

PR #150, opened 2020-03-16
Tests filters

pozylon:
Breaking:

- Full text search now correctly returns an empty resultset if the search has been upfront limited by a productId list of 0 array items (fixes an edge case with the local search provider)
- Unscoped Faceted search now returns all active filters when the local-search plugin is activated
- Unscoped Faceted search now returns even inactive filters if the filterQuery has a key of an existent filter

Minor:
- Allow skipInvalidation flags on createFilter and updateFilter


pozylon:
https://github.com/unchainedshop/unchained/issues/37

Issue #139, opened 2020-02-11
Inconsistent delivery & payment providers during complex cart operations


pozylon:
Mentioned by @claudiocro 

initCart currently sets order-delivery and order-payment, but the supportedDeliveryProviders can change when the cart changes, in those cases, order-delivery and order-payment are still set to providers that are invalid.

Consider some kind of logic that fixes the cart in those cases and maybe also add a possibility to let the developer configure which provider is going to be by default in which configuration.

pozylon:
The default resolver logic has been implemented with #153. It also allows to control the sort order based on the current order.

Idea 1:

Auto-fixxing invalid combinations is not possible at the moment as we can only control if an adapter is active for a specific configuration but we cannot control it on a more sophisticated "by-provider" level. Sooo...

1. We would have to provide a new custom function option to the modules replacing the newly added custom sort function, like:

```
const options = {
  modules: {
    delivery: {
      supportedProvidersForOrder: ({ order, providers }) => {
        if (hasBicycle(order)) return [BICYCLE_DELIVERY_PROVIDER_ID];
        return [DELIVERY_PROVIDER_ID,PICKUP_PROVIDER_ID]
      },
    },
    payment: {
      ...
    },
  }
};
```

2. Adjust the API in a way that checks if Order.payment.provider is still valid and return null for Order.payment (also do the same with delivery providers)

3. Because the first of the supported is set by unchained as a default provider on creation of a new cart, the UI could use the supportedPayment... API's also and trigger an updateCart when the situation occurs and null is returned as selected payment or delivery plugin.

PR #161, opened 2020-05-13
Supported providers

pozylon:
Fixes https://github.com/unchainedshop/unchained/issues/139


pozylon:
- It's now possible to control what delivery & payment providers are returned as supported depending on the order with a custom logic, that logic is run with every cart-mutating mutation (addCartProduct for ex.)
- the first in the returned order is the default which is set when a delivery/payment provider is not valid anymore or never has been valid (initialization of an order)

This should support https://git.panter.ch/manul/vps/veloplus-shop/-/issues/480

PR #174, opened 2020-05-23
Add order query tests

Mikearaya:
added order query test for:
- list of user orders
- error for anonymous user order query
- single order by id query


pozylon:
@Mikearaya nice!

PR #177, opened 2020-06-03
Mutation test round two

Mikearaya:
in this pull request, I have added tests for mutations

* `addAssortmentLink`
* `removeAssortmentLink`
* `reorderAssortmentLink`
* `removeAssortmentFilter`
* `reorderAssortmentFilter`
* `createFilterOption`
* `removeFilterOption`

I noticed `addAssortmentFilter` adds assortment filter without error when passed invalid filter ID, I don't think this is the desired behavior but for now, I have slipped the failing test. I will enable it if that is actually desired.


Mikearaya:
@pozylon  definitely.

PR #187, opened 2020-07-10
Controlpanel Enhancements

pozylon:


pozylon:
I've moved your cp changes to a branch "cp"

pozylon:
@Mikearaya I've been looking through this and I've seen that you've added custom CSS. Currently the whole controlpanel is more or less using default semantic ui react components (https://react.semantic-ui.com) so I was wondering what's the plan? Getting away from semantic? For example for tag items there are already 100'000 ways to do it with the semantic components (https://react.semantic-ui.com/elements/label/)

Mikearaya:
Yes I've used some CSS but they are not permanent, I was focusing on adding the functionality but I will move back to semantic when it's all done as moving away from it will add more problems than it solves.

PR #188, opened 2020-07-14
Exception handling

Mikearaya:
I have added few exception handling to make the API more descriptive when providing errors to failed requests such as
* when an item with the provided ID is not found either for the update, remove or read
* when invalid or null ID is provided for a request
in addition to the above and a few other refactors I have added inline comments on mutation API to make it understandable when calling a method from the playground


pozylon:
A huge PR! Thanks. Sometimes you just throw with new Error and sometimes you use one of the defined errors from the errors.js file. Also the controlpanel enhancements have sneaked into this PR. Can you please:

1. Move the CP stuff out (i think i‘ve prepared the cp PR for you to take over while you were offline)
2. Consistently use error types for all errors that are thrown in the resolvers so every error has a distinct error code and meta information when needed

packages/api/resolvers/mutations/createFilterOption.js
```diff
@@ -1,13 +1,18 @@
...
 import { Filters } from 'meteor/unchained:core-filters';
+import { FilterNotFoundError } from '../../errors';
 
 export default function (
   root,
   { option: inputData, filterId },
   { localeContext, userId },
 ) {
   log(`mutation createFilterOption ${filterId}`, { userId });
+  if (!filterId) throw new Error('Invalid filter ID provided');
```
pozylon, line 11: Here for example you "throw new Error(" instead of throw new InvalidFilterError

packages/api/resolvers/mutations/removeAssortmentFilter.js
```diff
@@ -1,11 +1,16 @@
...
 
 export default function (root, { assortmentFilterId }, { userId }) {
   log(`mutation removeAssortmentFilter ${assortmentFilterId}`, { userId });
+  if (!assortmentFilterId)
+    throw new Error('Invalid assortment filter ID provided');
   const assortmentFilter = AssortmentFilters.findOne({
     _id: assortmentFilterId,
   });
+  if (!assortmentFilter)
+    throw new AssortmentFilterNotFoundError({ assortmentFilterId });
```
pozylon, line 13: Good!

packages/api/schema/mutation.js
```diff
@@ -111,7 +113,10 @@ export default [
       ): OrderItem!
 
       """
-      Change billing address and order contact of an open order (cart)
+      Change billing address and order contact of an open order (cart). All of the parameters
+      except order ID are optional and the update will ocure for parameters provided.
+      If the delivery provider or payment provider ID provided doesn’t already exist new
+      providers will be created with the provided ID.
```
pozylon, line 119: New providers will be created? Sure?

packages/core-assortments/db/helpers.js
```diff
@@ -613,16 +614,21 @@ Collections.Assortments.helpers({
       this.parentIds(),
     );
   },
+  // returns AssortmentProducts and child assortment links with products.
```
pozylon, line 617: Good idea to comment this as the algo is pretty complex :)

packages/api/resolvers/mutations/makeQuotationProposal.js
```diff
@@ -11,6 +11,7 @@ export default function (
   { userId },
 ) {
   log('mutation makeQuotationProposal', { quotationId, userId });
+  if (!quotationId) throw new Error('Invalid quotation ID provided');
```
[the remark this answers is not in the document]
pozylon: "When invalid or null ID is provided for a request" -> this means that error would catch mutation mutate(requiredParameter: null) or mutate(requiredParameter: "") calls?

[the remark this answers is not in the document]
Mikearaya: i added this explicit error because i want to distinguish between not found because item ID doesn't exist and not found because the user passed invalid ID to give more clear error message.

packages/api/schema/mutation.js
Mikearaya: @pozylon https://github.com/unchainedshop/unchained/blob/master/packages/core-orders/db/orders/helpers.js#L594

Mikearaya: https://github.com/unchainedshop/unchained/blob/master/packages/core-orders/db/orders/helpers.js#L580

pozylon: Please fix this to "new order payments will be created", providers are never automatically created through a mutation.

pozylon:
Hey @Mikearaya can you please rebase and resolve so we can merge this?

packages/api/schema/mutation.js
Mikearaya: **Fixed**

Mikearaya:
@pozylon I'll try to remove all the control panel changes i made from this pull request by the end of this week.

Mikearaya:
@pozylon I've removed the panel changes

pozylon:
@Mikearaya Nope, still there

Mikearaya:
@pozylon I left the CSS classes intentionally because they are general bootstrap classes incase we might want to use them

pozylon:
So I guess the only thing that's left is 

- [ ] Introduction of a general IDInvalidError that you can use to throw when an ID is invalid (empty string, null) so that there is no string based "new Error" is left in the resolvers
- [ ] Rebase on master (fix tests run through)

pozylon:
Amazing you've added quite some stuff to this! I'll look into the test cases, thanks

PR #190, opened 2020-07-28
Add authors and convenience methods

pozylon:


Mikearaya:
@pozylon a little context would be helpful

pozylon:
:) The pr adds the "authorId" field to all db entities and sets it to the logged in user, this helps when we have to audit stuff

PR #191, opened 2020-07-31
Improve product search dropdown

harryadel:
@pozylon I applied changes according to your comments on this issue and #192. Now `selectImage` uses Semantic Icon as a fallback, media query only fetches a single element and the same component is used across the changes.

unchained1
unchained2
unchained3


pozylon:
Hey harry thank you for submitting this PR, looks good to me, except for the placeholder image, where i'd just use the image icon of semantic https://react.semantic-ui.com/elements/icon/

pozylon:
Also selectImage is ok, something more simple would be just using media(limit: 1) so 1 or 0 images would get returned

PR #192, opened 2020-08-03
Add search capability to assortment products select

harryadel:
This PR borrows from #191 and builds on top of it to provide a custom Uniforms Select field with search functionality for assortment products. But there're few things that needs to be tackled before finalizing this PR:

- `selectImage` function is also used in #191, so I reckon that it's best to export `selectImage` from a shared file for common use though what files? Should I just add a new file inside `lib` directory and export from it?

- When you load the assortment products page you actually get an error `Error: Cannot return null for non-nullable field AssortmentProduct.product.` which occurs at the query located in `packages/api/resolvers/queries/products.js`. The query fails to return any products because it's set to return products with active status only `i.e. -> { status: { '$eq': 'ACTIVE' } }`. Now, I don't really understand is this error happening because we're allowing attaching unactive products to an assortment (misconfigured display query) OR I've not properly populated my DB (Products should be set to active always?) [I'm not well versed in the application domain logic, so please bear with me]

unchained3

unchained2

### DIFF
--- a/examples/controlpanel/components/assortments/FormNewAssortmentProduct.jsx
+++ b/examples/controlpanel/components/assortments/FormNewAssortmentProduct.jsx
@@ -1,16 +1,124 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { toast } from 'react-toastify';
 import { compose, mapProps, withHandlers } from 'recompose';
-import { Segment } from 'semantic-ui-react';
+import {
+  Segment,
+  Dropdown,
+  Header,
+  Image as SemanticImage,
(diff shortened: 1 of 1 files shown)


pozylon:
Good catch, this should not happen ("Error: Cannot return null for non-nullable field AssortmentProduct.product.") and is an unchained bug we have to find and close in the backend

pozylon:
Furthermore, you should reuse the existing ProductDropdown you've made and maybe refactor it in a way that you can use it for multiple cases.

PR #196, opened 2020-08-05
Allow and display tags to assortment linking

harryadel:
### Linking Assortments to Filters
Thanks to the current state of code this PR was quite simple to implement, now you can sort link, add tags and display them along with the relationship.

To denote the parent child relationship a very simple tree view is implemented, please let me know if you'd like to change it.

assortmentslinkin


pozylon:
I just tested it and ordering and tags works as expected, thank you very much! I've found two things to improve:

1. it's hard to see that these items are orderable
2. the links when clicking on the title of an assortment do not work (the drag handler kicks in instead of route change)

Can we change the drag/drop behaviour for these lists in a way that we have an icon as active area to drag, something like this icon: https://semantic-ui.com/elements/icon.html#arrows -> arrows alternate

Also there is no <a> tag on the title of a product in assortment -> product link lists. So if you want to change that let's change it everywhere reordering is possible

harryadel:
Hey Pascal, thanks for your review. A drag handle has been added which fixed the problem of drag handler kicks in instead of route change and also the same was done for products sorting.

Issue #162, opened 2020-05-14
Enhance Control Panel


pozylon:
Extend Control Panel to easily search products & assortments, set tags for found items

- [x] New Search Popup to easily search for Products instead of simple Dropdowns that could grow huge
- [ ] Improve UX for displaying and Linking Assortments, Products & Filters
- [x] Show Breadcrumb data on Products
- [x] Show Tag data on Links, allow editing those
- [x] Visualize Taxonomies (Directed Graphs)

# New Search Popup & Improved Linking UX

Since the newest version we have a simple semantic-ui style search dropdown in Masterdata -> Products. That's cool because for some customers, it's just impossible to browse through the products without any filtering/search possibilities and only "load more".

- [x] Further improve the dropdown to show more details about a product, especially if it's a published product or not.

## Linking Products to Assortments

To link a product as child to an assortment, somebody has to go to the details of an assortment, select the vertical tab "Products" and then select the product to link. The current implementation has two big problems:

1. It's a simple select without the possibilitiy to filter/search
2. A query products(limit: 0) is sent to the server which takes forever to return when there are 20'000 products

- [x] Replace the select with a new dropdown equivalent to the new search dropdown to also improve finding the correct product to link and help with performance.

## Linking Products to a Configurable Product Variant

When using ConfigurableProduct, once can add variations and options in the vertical tab "Variations" where we also show some kind of select to choose a product from a list.

<img width="1401" alt="Bildschirmfoto 2020-06-11 um 13 33 16" src="https://user-images.githubusercontent.com/2109393/84380799-45cc4a00-abe8-11ea-8f1d-48953e9108fa.png">

- [x] Same as with assortment / product linking, reuse the search dropdown or go an equivalent way to enhance the UX and make it easier to link products to configurable products.

## Linking Assortments to Assortments

Introduce a new UX to easily link assortments to assortments, also improve the way children and parents are displayed and improve the sorting feature:

Problems with the current situation that need to be solved:
- [x] On the link list, it's not really clear what are parent and what are children relationships
- [x] Assortment to assortment links can be ordered freely and there is also a mutation for it but it is not leveraged in the control panel (a basic drag/drop sorting functionality is implemented in the assortment to product links tab as a reference)
- [x] AssortmentLinks not only link two assortments to each other, the have tags and meta, but the data is not displayed at the moment. at least tags should be visible.
```
query {
  assortment(slug: "schweiz") {
    linkedAssortments {
      _id
      sortKey
      child {_id}
[...]

PR #202, opened 2020-08-15
Directed graph

harryadel:
As discussed in #162, [react-cytoscapejs](https://github.com/plotly/react-cytoscapejs) is now used instead of the infinite table component to display assortments. I'd really appreciate your input on the looks of and code quality, thanks!

dag


harryadel:
ping @pozylon

pozylon:
Oh yeah will check it out today

pozylon:
Bildschirmfoto 2020-08-21 um 09 02 39

Just tried it with a backend with dozens of nodes, hit that error above, now going to try with a fresh copy

pozylon:
```
The above error occurred in the <t> component:
    in t (at CytoscapeComponent.jsx:56)
    in CytoscapeComponentWrapper (created by ForwardRef(LoadableComponent))
    in ForwardRef(LoadableComponent) (at AssortmentList.jsx:13)
    in AssortmentList (created by withHandlers(AssortmentList))
    in withHandlers(AssortmentList) (created by Mutation)
    in Mutation (created by Apollo(withHandlers(AssortmentList)))
    in Apollo(withHandlers(AssortmentList)) (created by Query)
    in Query (created by Apollo(Apollo(withHandlers(AssortmentList))))
    in Apollo(Apollo(withHandlers(AssortmentList))) (created by withState(Apollo(Apollo(withHandlers(AssortmentList)))))
    in withState(Apollo(Apollo(withHandlers(AssortmentList)))) (at assortments/index.jsx:12)
    in div (created by Segment)
    in Segment (created by TabPane)
    in TabPane (at assortments/index.jsx:11)
    in div (created by Tab)
    in Tab (at assortments/index.jsx:22)
    in div (created by Container)
    in Container (at assortments/index.jsx:20)
    in div (created by Segment)
    in Segment (at App.jsx:39)
    in main (at App.jsx:28)
    in App (created by mapProps(App))
    in mapProps(App) (created by Query)
    in Query (created by Apollo(mapProps(App)))
    in Apollo(mapProps(App)) (created by withRouter(Apollo(mapProps(App))))
    in withRouter(Apollo(mapProps(App))) (at assortments/index.jsx:19)
    in _c (at connectApollo.jsx:110)
    in ApolloProvider (at connectApollo.jsx:109)
    in WithData(_c) (created by App)
    in App
    in ErrorBoundary (created by ReactDevOverlay)
    in ReactDevOverlay (created by Container)
    in Container (created by AppContainer)
    in AppContainer
    in Root

React will try to recreate this component tree from scratch using the error boundary you provided, App.
```

pozylon:
Okay, one step further, now i get an empty window and I can't create a new assortment because that button is gone :D

harryadel:
@pozylon This is getting out of hand :D my bad for not checking things with fresh copy. I will recreate things locally and try out.

pozylon:
Oh wow you're extending the search to assortments? deep down the rabbit hole

harryadel:
To fix the white empty window issue, the same layout used in product list is replicated to have the table header and bottom, and to also have the search dropdown at the top so you can access the assortments. Search dropdown would've to be converted to a more of general component where depending on props it'd allow you to search products or assortments. More importantly, this introduces the need to implement assortment search capability just like product search which I'm having a bit of trouble with since assortments search returns nothing and I believe it has to do with its inability to find all assortments which gets filtered in later steps [I've started a review at the file which I think the problem lies within btw].

pozylon:
It's totally understandeable why you started to think in that direction but i guess it's maybe better to shortcut it for the moment and add a simple Query.searchAssortments

harryadel:
Hey Pascal

>It's totally understandable why you started to think in that direction but i guess it's maybe better to shortcut it for the moment and add a simple Query.searchAssortments

To be honest I didn't really get what you meant by implementing a simple search assortment query. :grimacing: Don't get me wrong it's not about discarding your words but lots of work has been implemented in the previous direction.  

The current search implementation requires passing `assortment: true` to differentiate the search (i.e products or assortments) which is later check for in `packages/core-filters/plugins/local-search.js` 

```
    if (query.assortment) {
...
```
As for the other issue, I'm thankful for the access you've given me though having the dump to be imported directly would allow debugging in much less time as with the GraphQL access manual search and import is required, maybe could be laziness on my side. Anyway, let me know what you think. Thanks!

pozylon:
Every time you invest lot of work in one direction you learn a lot about alternatives and inner workings, so I think that time is never wasted 🤟

Here are my thinkings:

When I designed the search concept the first time (we might have to re-think the current status quo in the future, though), I was never intending search to be used for assortments, as it started life as Assortment.search searching for products in a certain assortment. Later we have added Query.search which sounds very general but in fact should be named Query.searchProducts (because that's what it is).

Searching products comes with more stuff that is only important for products. The whole faceted filter implementation for example. That's why we now have all these parameters for Query.search which can be combined: 

- queryString (full-text search)
- filterQuery (faceted-filter search)
- assortmentId (search a specific assortment and it's sub-assortments)
- orderBy (order by)
- includeInactive (including inactive products)
- ignoreChildAssortments (opt-out of sub-assortment search)

Looking at this, the expected output should always be centralised around products. Assuming you have this configuration:

```
{
queryString: "Nike"
filterQuery: [{ key: "color", value: "black" }]
assortmentId: "assortment id of a assortment named socks"
}
```

And there is a assortment structure looking like this:
```
- Socks
-- Long Socks
-- Short Socks
- Hats
-- Cool Hats
-- Sport Hats
```

So this can be translated by "search for socks that are black in color inside the assortment socks and it's sub assortments, having a word Nike in either vendor, brand, title, subtitle,...."

I thought (because we're looking from a product search perspective):
When we find 3 socks with all of them beeing short socks, totalAssortments should be 2 (long & short) where as filteredAssortments should be 1 (short socks) and the assortments field should only return the short socks sub assortment

But:
Actually searching for assortments would translate to "search for assortments that are black in color inside the assortment socks and it's sub assortments, having a word Nike in either vendor, brand, title, subtitle,...." -> does not sound correct, right? Also it would not lead to useful result. 

I think this is one of the reasons why you added a flag?


I'd just make it very explicit by adding a Query.searchAssortments resolver that clearly state that "queryString" is a full text search of assortments. That also leaves open the opportunity of implementing the fields "totalAssortments, filteredAssortments, assortments" in the way I anticipated and let's us rename Query.search to Query.searchProducts in the future.

harryadel:
Wow, thanks a lot for the thorough explanation, I guess I need a little time to digest what you said. :sweat_smile:

packages/api/resolvers/types/assortment.js
```diff
@@ -5,7 +5,7 @@ export default {
   async assortmentPaths(obj, { forceLocale }, { localeContext }) {
     return obj.assortmentPaths(forceLocale || localeContext.normalized);
   },
-  async search(obj, query, context) {
```
pozylon, line 8: I'd just copy the new "searchProducts" resolver implementation and name it like the old:

async search -> obj.searchProducts
async searchProducts -> obj.searchProducts

packages/api/schema/types/search.js
```diff
@@ -5,13 +5,17 @@ export default [
     }
 
     """
-    Search result
+    Product search result
     """
-    type SearchResult {
```
pozylon, line 10: Also here, I'd just keep the old type with @deprecated

packages/core-filters/director.js
```diff
@@ -48,11 +54,20 @@ class FilterDirector {
     this.context = context;
   }
 
+  async buildAssortmentSelector(defaultSelector, options = {}) {
+    return this.reduceAdapters(async (lastSelector, concreteAdapter) => {
+      return concreteAdapter.transformProductSelector(
```
pozylon, line 59: This is probably wrong, .transformProductSelector in buildAssortmentSelector

packages/core-filters/director.js
```diff
@@ -65,6 +80,16 @@ class FilterDirector {
 
   async search(productIdResolver, options = {}) {
     return this.reduceAdapters(async (lastSearchPromise, concreteAdapter) => {
+      // concreteAdapter.context.query.assortmentIds
+      if (
+        concreteAdapter.__proto__.constructor.name === 'LocalSearch' &&
```
pozylon, line 85: It's not guaranteed that LocalSearch plugin is there and it's also not guaranteed that the user wants to use the LocalSearch plugin to search assortments

packages/core-filters/plugins/local-search.js
```diff
@@ -23,7 +28,7 @@ class LocalSearch extends FilterAdapter {
     return Boolean(context.query);
   }
 
-  async search(productIdResolver) {
+  async search(productIdResolver) { // Search Products
```
pozylon, line 31: A general search function like this looks like out of place

packages/core-filters/director.js
harryadel: I agree with your concerns but the alternative would be to pull out `searchAssortments` code into some other place and I thought it'd be nice to keep the search functionality in one place.

pozylon:
Can you please:
1. Add tests for Query.searchAssortments and also a "totalAssortments" Int field (helps for load more purposes when limit is set on assortments)
2. Please run "npm run test", it currently fails
3. The result type of searchProducts is still "SearchResult" but in fact it should be of type "ProductSearchResult" (api consistency)

harryadel:
-
- "totalAssortments" Int field has been [added](https://github.com/unchainedshop/unchained/pull/202/files#diff-f5294927685efdea824e95522bc6ae0eR5) along with tests for [Query.searchAssortments](https://github.com/unchainedshop/unchained/pull/202/files#diff-80262fa327b0e5397917ef969476dd6eR363).
- Fixed [broken tests](https://github.com/unchainedshop/unchained/pull/202/commits/3869cc9671d811b1ee83c9240f477d435bb69a02) but a weird thing happens when executing the entire test suite   `npx jest --runInBand` where some test fail randomly! Is We need to look into it :thinking: 
- Now, there's a reason for `search` [deprecation](https://github.com/unchainedshop/unchained/pull/202/commits/5df28fb41c4a95022ba36fa4faa77a61701ca781). BTW, should the [search directive](https://github.com/unchainedshop/unchained/pull/202/files#diff-685f258dad55bf02192e7c65c2ab7469R7) be removed? As I noticed that @deprecated was already used in UpdateEmail pre-adding my implementation.

pozylon:
@harryadel 

- Yes the deprecated directive can be removed as it is supported out of the box.
- Docs are good for the moment
- Tests pass

If I just call the query like this:
```
query {
  searchAssortments {
    totalAssortments
    assortments {
      _id
    }
  }
}
```

I get a weird error saying the assortmentId is not provided, that's maybe a copy-paste issue, I think it's totally fine to allow sending searchAssortments without a queryString

pozylon:
So with that change above the backend part LGTM

harryadel:
The copy-paste issue and directive removal have been implemented and as per our conversation via Basecamp I'd prefer to tackle the problem of dynamic child node click in a separate PR. So, for the time being the list/table view is re-added to allow for assortments viewership in flat way. And graph view is to be iterated upon in next PR.

PR #207, opened 2020-08-28
Api reference

pozylon:


pozylon:
Is this PR ready for review? I've rebased it and force-pushed to origin/api-reference so please hard-reset when working on it

Can you please:
- [ ] add a run script to package.json to build/serve the api reference
- [ ] update the changelog

Mikearaya:
Yes, it is, I was wondering when we were going to link it with the doc too. I'll update `package.json` with the necessary commands.

Mikearaya:
everything is ready, you need to have a running local instance of unchained before running this API generator.
after that use `npm run api-docs` inside `docs` folder

schmidsi:
I'll check it tomorrow

pozylon:
I've added a commit to remove the api folder from source control and also extend package with the installed @unchainedshop/dociql dependency

pozylon:
@Mikearaya coffeescript is a dependency, is this needed?

Mikearaya:
@pozylon no it's not needed, it works without it.

pozylon:
Also I tried something: When using static/api as destination folder and then npm run build (gatsby build) & npm run start (gatsby serve), the api will be available on http://localhost:9000 together with the gatsby docs page.

Mikearaya:
that's a good bonus for when someone wants to play around locally.

pozylon:
No it's actually very important to build on the CI.

Also, we want the CI to generate a Docker Image out of this docs. You could use "depends_on" (https://documentation.codeship.com/pro/builds-and-configuration/services-and-databases/) and actually run the introspection after the "minimal" service has started

PR #210, opened 2020-09-03
Bulk Import API

pozylon:
Test:
```
DEBUG=unchained:platform:bulk-import npx jest --watch --verbose --runInBand tests/bulk-importer.test.js
```

https://github.com/unchainedshop/unchained/issues/233

### DIFF
--- a/docs/source/api/entities.md
+++ b/docs/source/api/entities.md
@@ -1,0 +1,14 @@
+---
+title: "PIM Connector Framework"
+description: Import Data from any PIM
+---
+
+Entity: "Product"
+
+```json
+{
+  "entity": "PRODUCT",
+  "operation": "CREATE",
+  "payload": {}
+}
+```
--- a/docs/source/api/entities.md
+++ b/docs/source/api/entities.md
@@ -34,36 +34,43 @@ Set by unchained: authorId, slug history
       },
       "warehousing": {
         "baseUnit": "ST",
-        "sku": "",
+        "sku": ""
       },
       "supply": {
         "weightInGram": 0,
         "heightInMillimeters": 0,
         "lengthInMillimeters": 0,
-        "widthInMillimeters": 0,
+        "widthInMillimeters": 0
       },
       "proxy": {
         "assignments": [
           {
-            "vector": { "color": "red" },
+            "vector": {
+              "color": "red"
+            },
             "productId": "B"
           }
-        ],
+        ]
       },
       "plan": {
         "billingInterval": "daily",
         "billingIntervalCount": 1,
         "usageCalculationType": "metered",
         "trialInterval": "daily",
-        "trialIntervalCount": 1,
+        "trialIntervalCount": 1
       },
-      "bundleItems": [{
-        "productId": "c",
-        "quantity": 1,
-        "configuration": [
-          { "key": "greeting", "value": "For my Darling" }
-        ]
-      }],
+      "bundleItems": [
+        {
+          "productId": "c",
+          "quantity": 1,
+          "configuration": [
+            {
+              "key": "greeting",
+              "value": "For my Darling"
+            }
+          ]
+        }
+      ],
       "meta": {},
       "content": {
         "de": {
@@ -73,11 +80,29 @@ Set by unchained: authorId, slug history
           "slug": "produktname",
           "subtitle": "Kurzbeschreibung",
           "description": "Lange Beschreibung",
-          "labels": ["Neu"]
+          "labels": [
+            "Neu"
+          ]
         }
-      },
+      }
     },
-    "media": [],
+    "media": [
+      {
+        "asset": {
+          "_id": null,
+          "url": "https://www.story.one/media/images/poop-4108423_1920.width-1600.format-jpeg.jpg"
+        },
+        "sortKey": 0,
+        "tags": ["big"],
+        "meta": {},
+        "content": {
+          "de": {
+            "title": "Produktname",
+            "subtitle": "Kurzbeschreibung",
+          }
+        }
+      }
+    ],
     "variations": {
       "color": {
         "type": "COLOR",
@@ -88,7 +113,7 @@ Set by unchained: authorId, slug history
                 "title": "Rot",
                 "subtitle": ""
               }
-            },
+            }
           }
         },
         "content": {
@@ -96,9 +121,27 @@ Set by unchained: authorId, slug history
             "title": "Farbe",
             "subtitle": "Farbvariante"
           }
-        },
+        }
       }
-    }
+    },
+    "reviews": [
+      {
+        "_id": null,
+        "authorId": "root",
+        "rating": 1,
+        "title": "What the?",
+        "review": "That product just sucks big times",
+        "votes": [
+          {
+            "timestamp": "",
+            "userId": "root",
+            "type": "UPVOTE",
+            "meta": {}
+          }
+        ],
+        "meta": {}
+      }
+    ]
   }
 }
 ```
--- a/docs/source/api/entities.md
+++ b/docs/source/api/entities.md
@@ -29,6 +29,10 @@ Supported entity types:
 - PRODUCT
 - ASSORTMENT
 - FILTER
+- SUBSCRIPTION (coming soon)
+- ORDER (coming soon)
+- USER (coming soon)
+- REVIEWS (coming soon)
 
 Supported operations:
 - CREATE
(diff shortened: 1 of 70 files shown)


docs/source/api/entities.md
```diff
@@ -0,0 +1,329 @@
...
+
+### Endpoint
+
+The main entry point to send import events is:
+Mutation.sendBulkImportEvents. The mutation takes an array of events that consist out of a type, an operation and a payload.
+
+Supported entity types:
+- PRODUCT
+- ASSORTMENT
+- FILTER
```
schmidsi, line 31: Should we also provide the possibility to import users, orders and other stuff?

docs/source/api/entities.md
```diff
@@ -0,0 +1,329 @@
...
+
+```json
+{
+  "entity": "ENTITY TYPE",
+  "operation": "OPERATION TYPE",
+  "payload": { ... }
+}
+```
+
+Always try to send as many events at a time, so Unchained can optimize write operations.
```
schmidsi, line 48: There is probably an upper limit of how big one request can be. Would be good to find this limit and document it here.

docs/source/api/entities.md
```diff
@@ -0,0 +1,329 @@
...
+        "baseUnit": "ST",
+        "sku": ""
+      },
+      "supply": {
+        "weightInGram": 0,
+        "heightInMillimeters": 0,
+        "lengthInMillimeters": 0,
+        "widthInMillimeters": 0
+      },
+      "proxy": {
```
schmidsi, line 97: I do not understand what a proxy is by merely looking at this example. Is it explained elsewhere?

docs/source/api/entities.md
```diff
@@ -0,0 +1,329 @@
...
+    },
+    "media": [
+      {
+        "_id": null,
+        "created": null,
+        "updated": null,
+        "asset": {
+          "_id": null,
+          "contentHash": null,
+          "url": "https://www.story.one/media/images/poop-4108423_1920.width-1600.format-jpeg.jpg"
```
schmidsi, line 151: There should be the possibility to have media files bound to a language. E.g. A product-image might look different in German than in English. But it should be trivial to select the right one. Maybe add an optional field `language`?

docs/source/api/entities.md
```diff
@@ -0,0 +1,329 @@
...
+            {
+              "key": "greeting",
+              "value": "For my Darling"
+            }
+          ]
+        }
+      ],
+      "meta": {},
+      "content": {
+        "de": {
```
schmidsi, line 128: It would probably be better to use the longer lanuage code to be future proof? Like `de_CH`? Or accept both explicitly.

docs/source/api/entities.md
```diff
@@ -0,0 +1,329 @@
...
+          "de": {
+            "created": null,
+            "updated": null,
+            "title": "Farbe",
+            "subtitle": "Farbvariante"
+          }
+        }
+      }
+    ],
+    "reviews": [
```
schmidsi, line 195: Reviews should probably be in a separate entity because there will be a lot of appending operations (e.g. add a review to a product) which could probably slow down the import if they are all updating the product itself.

docs/source/api/entities.md
```diff
@@ -0,0 +1,329 @@
...
+  "operation": "CREATE",
+  "payload": {
+    "specification": {
+      "_id": "A",
+      "created": null,
+      "updated": null,
+      "sequence": "",
+      "isActive": true,
+      "isBase": false,
+      "isRoot": true,
```
schmidsi, line 238: Difference between isBase and isRoot is not obvious for me.

docs/source/api/entities.md
```diff
@@ -0,0 +1,329 @@
...
+- authorId,
+- slug history
+- _id, contentHash, created & updated if not provided
+
+```json
+{
+  "entity": "PRODUCT",
+  "operation": "CREATE",
+  "payload": {
+    "specification": {
```
schmidsi, line 64: What other top-level fields could `payload` have than `specification`? Why not remove `specification` and directly add the data?

schmidsi (commented):
Looks great! I just added some remarks.

Mikearaya:
the structure looks good to me, I have a question though how is the `_id` generated meaning, is it going to be the same as the ID from the  PIM or ERP or is it autogenerated on unchained. I'm asking this because I'm thinking how update and remove events will be handled.

docs/source/api/entities.md
pozylon: For some cases it might be beneficial to additionally import users, orders, reviews and subscriptions. Especially if somebody migrates from an existing webshop. I will add it and specify that this will be v2 of the Import API

docs/source/api/entities.md
pozylon: Good catch. That is when a product "proxies" other products, which is the case with ConfigurableProducts and BundleProducts which when you add them to a cart, resolve to actual SKU's by some algorithm. That could also be the case for "Alias Products" or "Surrogate Products" we have discussed a while ago.

But currently, the BundleProduct uses the bundleItems data so it's maybe better to call this "variantResolvers" and skip proxy.assignments

docs/source/api/entities.md
pozylon: This is totally unsupported by unchained (and you would have to solve it by introduction of "lang" tags or something) at the moment so I will skip this

docs/source/api/entities.md
pozylon: There is no rules on how languages can be added so both are accepted, you could add a Language with an isoCode of "klingon" and use that here.

docs/source/api/entities.md
pozylon: I once wanted to change the default behavior of Query.assortment in a way that when you did not specify assortmentId it will return the base assortment. This would solve some issues we had at different clients where we had include/exclude certain assortments when just going for the isRoot ones.

pozylon: what do you think?

docs/source/api/entities.md
pozylon: Currently, I've designed Product and Assortment entity types in a way that relationships and assets are part of them, so there is payload.specification, payload.media & payload.variations. The sync will only update the parts where the toplevel field is set. Alternatively I could use entities like PRODUCT_MEDIA and PRODUCT_VARIATION but in those cases the sync microservice will have to send the correct REMOVE operations for these entities which is overkill I guess... Have to think about it again

docs/source/api/entities.md
schmidsi: Right, tags can solve this.

docs/source/api/entities.md
schmidsi: That makes sense.

docs/source/api/entities.md
schmidsi: Hmm, I don't get it. Can you elaborate the case more, where you want to have a `isBase: false` but `isRoot: true`? Maybe it's a good learn for others too.

pozylon:
> the structure looks good to me, I have a question though how is the `id` generated? meaning, is it going to be the same as the ID from the PIM or ERP or is it autogenerated on unchained. I'm asking this because I'm thinking how update and remove events will be handled.

Good question, I think it should be the same as the one that comes from the PIM/ERP. If there is no unique ID coming from the PIM/ERP, you have to create one by hashing some kind of data you can refer to later (in the microservice). Also it's the microservice that needs to find out if it needs to send Update or Create or Delete calls but it's totally fine for the microservice to use the GraphQL API to find that out

docs/source/api/entities.md
pozylon: isRoot assortments in fictional shop: Public Products (isBase = true), Mandate Y Products, Mandate X Products
Public Products child assortments: Men, Women, Children

Requirement: Only show public products

pozylon:
npx jest --watch --verbose --runInBand tests/bulk-importer.test.js

pozylon:
Some remarks: The current implementation pretty much trusts the admin user that pushes in data, only very rudimentary simple schema based validations are done. For a later version, we will need a fine grained check for all operations so nobody can add a product with a type of "NotExistingProductType" for example.

Also, although I prepared a solution to sum up all MongoDB operations in so-called "Ordered Bulk Operations", I've implemented the actual logic in a stupid way that does not use that concept yet.

So I consider this a beta version but I don't want to spend too much time perfectioning it at the moment.

PR #212, opened 2020-09-09
Add nodes on-click 

harryadel:
This PR relies on `cy` prop to set up event listeners to invoke the fetch and display logic for linked assortments of a node(assortment). 

The two event listeners are:
- `cy.on 'tap', 'node'`
This event listens to clicks/taps on nodes and then fetches the linked assortments using Apollo client:
```
const { data } = await client.query({
            query: GET_ASSORTMENT_LINKS,
            variables: { assortmentId: node.id() },
          });
```

- `cy.on 'add', node`
```
cy.on('add', 'node', () => {
          cy.layout(layout).run();
          cy.fit();
        });
```
The listens to the the addition of any new nodes and invokes(resets) the layout to avoid stacking the nodes on top of each other.

### DIFF
--- a/examples/controlpanel/components/assortments/AssortmentGraph.jsx
+++ b/examples/controlpanel/components/assortments/AssortmentGraph.jsx
@@ -47,7 +47,7 @@ const AssortmentGraph = ({ data, router }) => (
     <Table.Body>
       <Table.Row>
         <Table.Cell>
-          <CytoscapeComponentWithNoSSR data={data} />
+          <CytoscapeComponentWithNoSSR assortments={data.assortments} />
         </Table.Cell>
       </Table.Row>
     </Table.Body>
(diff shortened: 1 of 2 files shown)


pozylon:
Tested it a bit, ran into a problem:

A (root) > B (root) > C
A (root) > C

I can click on B and it will show C behind B but there is no arrow from A to C then. If I refresh the page and click on A it will show C but there will be no arrow from B to C.

I guess you're not evaluating the parent links for newly added items

harryadel:
@pozylon Please recheck, ty.

pozylon:
@harryadel Now looks good when testing locally but leads to the real life scenario on the Bitbook GraphQL endpoint

harryadel:
:confetti_ball:

PR #220, opened 2020-09-11
Assortment to assortment linking dropdown

harryadel:
This PR replaces the dropdown at assortments linking with our search dropdown component. Also, the search component was modified to look as a normal uniforms component so now it turns red when form submission fails and what not.   

Pascal_1
Pascal_2


pozylon:
Nice work!

PR #221, opened 2020-09-12
Add tags to assortment filters

harryadel:
Give the assortment filters form a similar look to assortment linking.

Pascal


pozylon:
LGTM

Issue #216, opened 2020-09-09
Get rid of "aldeed:schema-index"


pozylon:
It's not the job of a schema validator or collection2 to automatically create indexes based on a flax index: true.

- [x] Remove all traces of aldeed:schema-index
- [x] Change all schemas that contain index: true and add explicit index creating code for all collections that had index: true flags. See the core-worker/db/schema.js file for example

Mikearaya:
@pozylon default sort order will be ascending? if I'm correct.

pozylon:
Nice, first part of meteor purge is done

PR #222, opened 2020-09-13
Aldeed index replacement #216

Mikearaya:
Remove all traces of `aldeed:schema-index` with explicit `createIndex` definition

### DIFF
--- a/packages/core-subscriptions/db/subscriptions/schema.js
+++ b/packages/core-subscriptions/db/subscriptions/schema.js
@@ -25,9 +25,9 @@ export const PeriodSchema = new SimpleSchema(
 
 export const Schema = new SimpleSchema(
   {
-    userId: { type: String, required: true, index: true },
-    status: { type: String, required: true, index: true },
-    productId: { type: String, required: true, index: true },
+    userId: { type: String, required: true },
+    status: { type: String, required: true },
+    productId: { type: String, required: true },
     quantity: { type: Number },
     configuration: Array,
     'configuration.$': {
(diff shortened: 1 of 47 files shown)


packages/core-subscriptions/db/subscriptions/schema.js
```diff
@@ -64,3 +64,7 @@ export const Schema = new SimpleSchema(
 );
 
 Subscriptions.attachSchema(Schema);
+
+Subscriptions.rawCollection().createIndex({ userId: 1 });
```
pozylon, line 68: Please create the indexes with an "init" function like in other packages (core-warehousing for ex.) so we can control the order of execution and it's consistent

pozylon (changes requested):
Please see my comment about the init function.  Also in other packages this has to be changed.


What's more is that I think there is a lot of migrations missing. For example in core-bookmarks I don't see any code that drops and recreates the index, which will lead to duplicate indexes with different names.

Mikearaya:
@pozylon I did that by first assigning false to index before using create index to create the index. That's why. If you saw on assortment I added the migration code but then I thought that would be duplicate. 

Sure I'll add it

pozylon:
I'm afraid that branch looks totally messed up. There are commits in the tree that look like they are cherrypicked or merged without merge commit or something because they already exist and have different commit hashes (0c7b99fd5424dac073c98f1caa70d170b954665e, a9fc79ef4e4334f3aabee9dfadeba762d7bfdd63). If I'd merge that, commits be duplicate.

Before you request review, please always make sure that:

1. The branch is rebased from master and has a linear history on top of master, if you have problems with mixed up trees, try a gui client like https://www.gitkraken.com, 
2. The branch is mergable to master without conflicts
3. The files changed tab in Github cannot contain any change that is not part of your work (https://github.com/unchainedshop/unchained/pull/222/files)

Mikearaya:
@pozylon this is all happening when i rebase. ok i'll create another branch

PR #227, opened 2020-09-15
Themable CP

pozylon:


pozylon:
With this branch, CP can be themed by following https://semantic-ui.com/usage/theming.html

schmidsi (approved):

schmidsi:
So this chance allows us to theme the CP globally, but not per customer, right?

pozylon:
Yeah, just globally, but a customer could copy the controlpanel folder over to his project and have it themed

pozylon:
Das sind alli global variables womer cha setzä:

https://github.com/Semantic-Org/example-github/blob/master/semantic/src/themes/default/globals/site.variables

Issue #226, opened 2020-09-14
Possibility to search for an user in the Controlpanel


schmidsi:
**Introduction**
As an admin I want to be able to search for users by typing their name in order to quickly find the user and inspect its current state.

**Is your feature request related to a problem? Please describe.**
Currently, the `users(...)` query is not able to receive search params

**Describe the solution you'd like**
It would be possible without changing the GraphQL Query with a client-side in-memory search but that solution would not scale.

**Describe the design of the solution in detail**
Extend users-query:
```graphql
users(
limit: Int = 20
offset: Int = 0
includeGuests: Boolean = false
queryString: String # <- new
): [User!]!
```

This query string would result in a simple MongoDB query in the fields: `username`, `name`, `emails` and maybe also `userProfile`.

Then, use this extended query to have search field in the controlpanel on top of the users-list. Typing in this field would filter the userlist below.

pozylon:
@schmidsi The frontend part should also be part of the solution, right?

schmidsi:
Yes

PR #228, opened 2020-09-15
Add queryString to users query

harryadel:
Added search functionality to users query
https://github.com/unchainedshop/unchained/issues/226

### DIFF
--- a/packages/api/resolvers/queries/users.js
+++ b/packages/api/resolvers/queries/users.js
@@ -1,9 +1,20 @@
 import { log } from 'meteor/unchained:core-logger';
 import { Users } from 'meteor/unchained:core-users';
 
+Users.rawCollection().createIndex({
+  username: 'text',
+  'emails.$.address': 'text',
+  'profile.displayName': 'text',
+  'lastBillingAddress.firstName': 'text',
+  'lastBillingAddress.lastName': 'text',
+  'lastBillingAddress.company': 'text',
+  'lastBillingAddress.addressLine': 'text',
+  'lastBillingAddress.addressLine2': 'text',
+});
+
 export default function users(
   root,
-  { limit, offset, includeGuests },
+  { limit, offset, includeGuests, queryString },
   { userId },
 ) {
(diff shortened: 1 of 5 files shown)


packages/api/resolvers/queries/users.js
```diff
@@ -1,19 +1,38 @@
...
+  { limit, offset, includeGuests, queryString },
   { userId },
 ) {
   log(
     `query users ${limit} ${offset} ${includeGuests ? 'includeGuests' : ''}`,
     { userId },
   );
 
+  if (queryString) {
+    return Users.find(
```
pozylon, line 26: Shouldn't we allow combining includeGuests with the queryString on the backend?

packages/api/resolvers/queries/users.js
```diff
@@ -1,19 +1,38 @@
 import { log } from 'meteor/unchained:core-logger';
 import { Users } from 'meteor/unchained:core-users';
 
+Users.rawCollection().createIndex({
```
pozylon, line 4: Please move this to the initializing function which also starts the migrations (see upcoming changes in https://github.com/unchainedshop/unchained/pull/229 for reference)

tests/auth-admin.test.js
```diff
@@ -19,9 +19,15 @@ describe('Auth for admin users', () => {
   describe('Query.users', () => {
     beforeAll(async () => {
       const Users = db.collection('users');
+
+      Users.createIndex({
```
pozylon, line 23: Tests should not create indexes, I checked and I've seen that there is also a createIndex in the products seeds but I don't get why, maybe the important thing is awaiting it?

pozylon (changes requested):

tests/auth-admin.test.js
harryadel: Because there're tests that utilize the search functionality and they don't seem to work unless we set the indexes.

pozylon:
LGTM

PR #229, opened 2020-09-15
Aldeed index removal #216

Mikearaya:


pozylon:
Thanks, this looks good but I don't know how to test if this all works :)

pozylon:
I just tried to prepare a legacy unchained system and then checkout the branch and all the migrations went through, I think we can only really see if there is any index missing when testing with a system that is actually in use.

Mikearaya:
@pozylon I'm 99.9% sure I've replaced all the previous indexes :).

And yes compound index will be after, identifying which ones not so much :). I think the best approach would be to keep that in mind and add the modification to each package when we see it.

pozylon:
Nope, sorry, no merge yet by speartesting this with product media:

db.getCollection("product_media").getIndexes() -> only _id, should return the indexes defined here https://github.com/unchainedshop/unchained/pull/229/files#diff-fab070208ba4987245369bed43afbbd2R52

This code sadly is not tested at all

Mikearaya:
@pozylon sorry about that i assumed all the schema default exports in every package were were were executed, which was the issue. some files were not.
on the positive side, I've checked if all the assigned index existed and they all do. so that solves your previous concern. sorry for the inconvenience but now everything is ready for merge.

Mikearaya:
@harryadel i think the introduction of index on user `email.$.address` is not valid. 
https://github.com/unchainedshop/unchained/pull/228/files#diff-07b0240dd1bfa1e2859eeeff0267e02bR183

so if it's not necessary we can remove it or find a better way to assign it as index.

pozylon (approved):

Issue #225, opened 2020-09-14
Get rid of Meteor Accounts


pozylon:
Alternative frameworks we have looked into:
- Accounts JS https://github.com/accounts-js/accounts (thanks @harryadel )
- Ooth https://github.com/nmaro/ooth (maybe fork?)
- Passport (http://blog.thebakery.io/meteor-accounts-auth-with-passportjs/)

Decision: Accounts JS (maintained and modern)

Requirements for Meteor Accounts Hop-Off / Purge:

- [ ] Existing cookies / tokens should still work after migration, so that we don't have to invalidate all user sessions when upgrading a client, luckily accountsjs seems flexible enough (https://www.accountsjs.com/docs/api/server/interfaces/accountsserveroptions#optional-createnewsessiontokenonrefresh, https://www.accountsjs.com/docs/api/server/interfaces/accountsserveroptions#optional-tokencreator)

- [x] Is 2FA ready and supports different strategies

- [ ] Easy way to extend user fields

- [ ] Documented migration path for custom login handlers and events (Accounts.registerLoginHandler, Accounts.onLogin) (https://github.com/meteor/meteor/blob/devel/packages/accounts-base/accounts_server.js) to Accounts.js as we currently use these features for different clients

pozylon:
First we will need some kind of migration path :)

harryadel:
If it's ok with you I may start tinkering and see where it leads.

pozylon:
Sure thing, thanks

harryadel:
The plan is as follows:

[] Tackle login and user-creation - the corner stone of any working authentication module and also this ensures we don't break the authentication for current users.

[] Replace the other functions like password reset, password change etc. This could be split up into PRs with each one only undertaking one function. Instead of major PR containing all of the changes. 

[] Remove Meteor accounts/base package and test for any errors.

It's important to note that with each change the associated hooks should be changed as well.

pozylon:
Nice, just one remark: I‘d go one step further and define milestone 1 to be login only. So user creation still through meteor accounts. Then you could proof that login with legacy hashes work

stolinski:
This might be an odd question for this repo considering I'm not using unchained. But have you had luck with accounts.js logging into Meteor accounts? I have a very similar setup to your accounts branch, but am not getting creds to validate. Sorry for making an off topic post in your repo, but I'd love some help with my implementation as it's VERY similar to yours.

harryadel:
[...]

Issue #217, opened 2020-09-09
Get rid of "meteor/unchained:roles"


pozylon:
The current roles package is a fork of nicolaslopezj:roles and depends on a lot of legacy code written in ES5, using underscore (the meteor version) and generally heavily depends on Meteor specifics.

- [x] Tests merged in master, enhanced with permission checks as @Mikearaya described here: https://github.com/unchainedshop/unchained/issues/189
- [x] Remove the package
- [x] Completely rebuild from scratch (in TS) and put into meteor/unchained:api package so that you can use it as a drop-in replacement in packages/api/roles/index.js

Issue #169, opened 2020-05-19
Support a different name for cookie based token auth


pozylon:
https://github.com/unchainedshop/unchained/blob/a842e5d1321cf6e1319b95f4d42a22dec025a226/packages/api/user-context.js#L7

It should be more something different like "unchained_token" or maybe even customizable rather than meteor_login_token

pozylon:
@toinhao1 Or this

pozylon:
Already done and uses "token"

PR #230, opened 2020-09-16
Migrate off of Meteor Accounts

harryadel:
This PR is intended for https://github.com/unchainedshop/unchained/issues/225. My plan is to slowly modify each file inside of `packages/api/resolvers/mutations/accounts` with no changes to the schema whatsoever to avoid breaking client's code. Once we're done with this, https://github.com/unchainedshop/unchained/issues/217 is next as it's intertwined in a sense and so is https://github.com/unchainedshop/unchained/issues/169.


pozylon:
https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_omit

packages/api/resolvers/mutations/accounts/createUser.js
```diff
@@ -12,5 +17,15 @@ export default async function createUser(root, options, context) {
     mappedOptions.password = hashPassword(mappedOptions.plainPassword);
     delete mappedOptions.plainPassword;
   }
-  return callMethod(context, 'createUser', mappedOptions);
+  const userId = await accountsPassword.createUser(mappedOptions);
```
pozylon, line 19: If context is not submitted to createUser, it will be missing in any hooks regarding the createUser event, no?

packages/api/resolvers/mutations/accounts/loginAsGuest.js
```diff
@@ -1,10 +1,16 @@
...
+export default async function loginAsGuest(root, methodArguments, context) {
   log('mutation loginAsGuest');
-  return callMethod(context, 'login', {
-    ...methodArguments,
-    createGuest: true,
-  });
+  const {
+    user: { userId },
+    token,
+  } = await accountsServer.loginWithService('guest', methodArguments, context);
```
pozylon, line 9: Inconsistently here the context is sent to the accountsserver method but it was not for changePassword and createUser

packages/core-accountsjs/core-accountsjs.js
```diff
@@ -0,0 +1,203 @@
...
+  EmailAlreadyExists: 'EmailAlreadyExists',
+  /**
+   * Username already exist in the database.
+   */
+  UsernameAlreadyExists: 'UsernameAlreadyExists',
+};
+
+const METEOR_ID_LENGTH = 17;
+
+export const idProvider = () =>
```
pozylon, line 41: Why did you replicate the id generation?

I think it's not important that the _ids are created in the same way as Meteor because we don't have to create the id's on the client side, the default for MongoDB is propably a  better choice (ObjectId).

packages/core-accountsjs/core-accountsjs.js
```diff
@@ -0,0 +1,203 @@
+import { AccountsServer, ServerHooks, AccountsJsError } from '@accounts/server';
+import { AccountsPassword } from '@accounts/password';
+import MongoDBInterface from '@accounts/mongo';
+import { MongoInternals } from 'meteor/mongo';
+import { DatabaseManager } from '@accounts/database-manager';
+import { Random } from 'meteor/random';
+import crypto, { randomBytes } from 'crypto';
+import pick from 'lodash.pick';
```
pozylon: I'm opposed to adding any dependencies that are not needed for important reasons, check out https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_pick and I think you don't need to use defer either in a strictly node.js environment (https://medium.com/javascript-in-plain-english/setimmediate-vs-process-nexttick-in-nodejs-6fbfbebd60a3)

packages/api/package.js
```diff
@@ -8,7 +8,6 @@ Package.describe({
 
 Npm.depends({
   accounting: '0.4.1',
-  'lodash.omit': '4.5.0',
```
pozylon: You have a good chance to keep your package free from any dependencies except for accountsjs, see my comments about lodash in the review

examples/minimal/package.json
```diff
@@ -31,6 +31,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.11.2",
+    "@graphql-modules/core": "^0.7.17",
```
pozylon, line 34: This should be part of Npm.depends i guess

packages/platform/setup-accounts.js
```diff
@@ -82,29 +92,28 @@ export default ({ mergeUserCartsOnLogin = true } = {}) => {
...
 
-  Accounts.onLogin(({ methodArguments, user }) => {
+  accountsServer.on('LoginSuccess', ({ user, connection = {} }) => {
     const {
       userIdBeforeLogin,
       countryContext,
       remoteAddress,
       normalizedLocale,
-    } = [...methodArguments].pop() || {};
+    } = connection;
```
pozylon, line 116: I guess these parameters are not part of the connection, these are parameters form the GraphQL context, sure that Cart Merging works they way it should?

pozylon (commented):
I think you are on a very good track. I only concentrated on the overall architecture and did not check functionality yet.

pozylon:
```
[dev:minimal     ] W20200924-13:00:07.125(2)? (STDERR) (node:29021) UnhandledPromiseRejectionWarning: Error: Meteor code must always run within a Fiber. Try wrapping callbacks that you pass to non-Meteor libraries with Meteor.bindEnvironment.
[dev:minimal     ] W20200924-13:00:07.125(2)? (STDERR)     at Object.Meteor._nodeCodeMustBeInFiber (packages/meteor.js:1186:11)
[dev:minimal     ] W20200924-13:00:07.125(2)? (STDERR)     at Object.Meteor.bindEnvironment (packages/meteor.js:1260:10)
[dev:minimal     ] W20200924-13:00:07.125(2)? (STDERR)     at MongoConnection.<anonymous> (packages/meteor.js:305:25)
[dev:minimal     ] W20200924-13:00:07.125(2)? (STDERR)     at MongoConnection.<computed> [as update] (packages/mongo/mongo_driver.js:785:49)
[dev:minimal     ] W20200924-13:00:07.126(2)? (STDERR)     at Collection.update (packages/mongo/collection.js:588:31)
[dev:minimal     ] W20200924-13:00:07.126(2)? (STDERR)     at Collection.Mongo.Collection.<computed> [as update] (packages/aldeed:collection2/collection2.js:214:19)
[dev:minimal     ] W20200924-13:00:07.126(2)? (STDERR)     at Collection.Users.updateHeartbeat (packages/unchained:core-users/db/helpers.js:226:9)
[dev:minimal     ] W20200924-13:00:07.126(2)? (STDERR)     at packages/unchained:platform/setup-accounts.js:118:11
[dev:minimal     ] W20200924-13:00:07.126(2)? (STDERR)     at /Users/pozylon/Repositories/unchained/examples/minimal/.meteor/local/isopacks/unchained_core-accountsjs/npm/node_modules/emittery/index.js:227:13
[dev:minimal     ] W20200924-13:00:07.126(2)? (STDERR)     at Array.map (<anonymous>)
[dev:minimal     ] W20200924-13:00:07.126(2)? (STDERR)     at Emittery.emit (/Users/pozylon/Repositories/unchained/examples/minimal/.meteor/local/isopacks/unchained_core-accountsjs/npm/node_modules/emittery/index.js:225:23)
[dev:minimal     ] W20200924-13:00:07.126(2)? (STDERR)     at processTicksAndRejections (internal/process/task_queues.js:97:5)
[dev:minimal     ] W20200924-13:00:07.127(2)? (STDERR) (node:29021) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 3)
[dev:minimal     ] W20200924-13:00:07.127(2)? (STDERR) (node:29021) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

packages/api/resolvers/mutations/accounts/loginAsGuest.js
harryadel: [LoginWithService](https://www.accountsjs.com/docs/api/client/interfaces/transportinterface#loginwithservice) utilizes a different api from [createUser](https://www.accountsjs.com/docs/api/password/classes/accountspassword/#createuser); that's why one of them gets the context as an additional parameter and no the other. As for the hooks, it doesn't really affect us so far, maybe if you'd like the context data to be passed that's a different story.

packages/core-accountsjs/core-accountsjs.js
harryadel: Early on during the implementation https://github.com/lorensr/apollo-accounts was used a guidance and I wanted to play it as safe as possible by sticking to Meteor's way of doing things. Though, if it's ok with you we can surely change it.

packages/core-accountsjs/core-accountsjs.js
harryadel: The Lodash part is already done. I'mma see what can be done now about deferring...

pozylon:
Functionally the most important parts seem to work just fine:
- Login with exsting hash
- Changing the password of existing E-Mail
- Registering as new user

But:
- Guest Login is broken, if I use the loginAsGuest mutation and then use the token as Bearer token I cannot get my user _id

Just some things that I encountered but are not to be fixed in this PR:

- Currently it's possible to add an E-Mail address without suffix but it's not possible to create a new user without a suffix
- The account enrollment currently does not trigger an enrollment E-Mail when an admin adds a new user through CP

examples/minimal/package.json
harryadel: Great attention to details as always.

packages/platform/setup-accounts.js
harryadel: >Cart Merging works they way it should
:grin: let me check.

pozylon: If not, please add a test along the way that checks if a cart is actually migrated. The idea is:

1. Login as guest
2. Put Product X in the cart
3. Login as not-guest
4. Check if logged in as the not-guest user and the cart contains Product X

pozylon: It's really important for cases where you are asked to login during the checkout

packages/core-accountsjs/core-accountsjs.js
pozylon: Ok

packages/api/resolvers/mutations/accounts/loginAsGuest.js
```diff
@@ -1,10 +1,17 @@
...
+import { accountsServer } from 'meteor/unchained:core-accountsjs';
 
-export default function loginAsGuest(root, methodArguments, context) {
+export default async function loginAsGuest(root, methodArguments, context) {
   log('mutation loginAsGuest');
-  return callMethod(context, 'login', {
-    ...methodArguments,
-    createGuest: true,
-  });
+  const { user: id, token } = await accountsServer.loginWithService(
```
pozylon, line 6: In the loginWithPassword resolver you use evaluateContext and here you don't?

packages/api/resolvers/mutations/accounts/createUser.js
pozylon: Still the case? Also for changePassword and I guess you should pass it with your evaluateContext function

packages/core-accountsjs/core-accountsjs.js
```diff
@@ -0,0 +1,205 @@
...
+  },
+);
+
+const dbManager = new DatabaseManager({
+  sessionStorage: mongoStorage,
+  userStorage: mongoStorage,
+});
+
+const accountsServerOptions = {
+  siteUrl: Meteor.absoluteUrl(),
```
pozylon, line 65: Please don't introduce new code dependencies by using Random or Meteor in this package 😀

packages/core-accountsjs/core-accountsjs.js
```diff
@@ -0,0 +1,205 @@
...
+      throw e;
+    }
+  }
+}
+
+export const accountsPassword = new UnchainedAccountsPassword({});
+
+class UnchainedAccountsServer extends AccountsServer {
+  destroyToken = (userId, loginToken) => {
+    this.users.update(userId, {
```
pozylon, line 149: Did you test if destroyToken works?

tests/cart-migrate.test.js
```diff
@@ -0,0 +1,126 @@
...
+            }
+          }
+        }
+      `,
+    });
+    // Although the admin has placed no orders you can find an order of his
+    const adminOrders = await db
+      .collection('orders')
+      .findOne({ userId: loginWithPassword.id });
+    expect(adminOrders).toMatchObject({});
```
pozylon, line 124: This will always be true, I guess? You should check for the same orderId that was used in result.data.addCartProduct

tests/helpers.js
```diff
@@ -42,7 +42,8 @@ export const setupDatabase = async () => {
   await seedDeliveries(db);
   await seedPayments(db);
   await seedWarehousings(db);
-  await seedOrders(db);
+  // eslint-disable-next-line no-unused-expressions
+  !args?.skipOrders && (await seedOrders(db));
```
pozylon, line 46: Please remove this change

pozylon (commented):
I just went through it but did not check for functionality yet.

packages/platform/setup-accounts.js
```diff
@@ -21,22 +24,31 @@ export const buildContext = (user) => {
...
+      .extend({
+        password: String,
+      })
+      .omit('created')
+      .validate(clone);
+    return user;
+  };
 
-  Accounts.onCreateUser((options = {}, user = {}) => {
+  accountsServer.on('CreateUserSuccess', async ({ user, options }) => {
```
pozylon, line 48: I think CreateUserSuccess is not the equivalent of Accounts.onCreateUser, maybe it has to be moved to  and I guess the google and facebook related code has to be removed as there is own strategies for that in accounts-js.

Returning a user from the event will not do anything, so changes to newUser will be going to nirvana.

And I'm not sure that skipEmailVerification option does what it should

packages/platform/setup-accounts.js
```diff
@@ -56,16 +67,15 @@ export default ({ mergeUserCartsOnLogin = true } = {}) => {
...
+    if (!newUser.guest && !options?.skipEmailVerification) {
+      accountsPassword.sendVerificationEmail(
+        Users.findOne({ _id: user._id }).primaryEmail()?.address,
+      );
     }
     return newUser;
   });
 
-  Accounts.removeOldGuests = (before) => {
+  accountsServer.removeOldGuests = (before) => {
```
pozylon, line 78: I think you can either remove this for now or add an admin mutation that allows calling that method.

packages/platform/setup-accounts.js
```diff
@@ -82,29 +92,27 @@ export default ({ mergeUserCartsOnLogin = true } = {}) => {
...
-      return undefined;
-    }
-    const guestOptions = createGuestOptions(options.email);
-    return {
-      userId: Accounts.createUser(guestOptions),
-    };
-  });
+  accountsServer.services.guest = {
+    async authenticate(params) {
+      const guestOptions = createGuestOptions(params.email);
```
pozylon, line 103: If you throw here it is the equivalent of throwing in accountsServer.on('ValidateLogin' so I guess you could move the code in ValidateLogin to here and make it more clear what happens when it's a guest login

packages/platform/setup-accounts.js
pozylon: Also I think the closest thing to Accounts.onCreateUser is if you use this function here to adjust the user fields: https://www.accountsjs.com/docs/api/server/interfaces/accountsserveroptions#optional-userobjectsanitizer

packages/core-accountsjs/core-accountsjs.js
```diff
@@ -0,0 +1,205 @@
...
+  userStorage: mongoStorage,
+});
+
+const accountsServerOptions = {
+  siteUrl: Meteor.absoluteUrl(),
+  // sendVerificationEmailAfterSignup: false, // It's set to false by default
+};
+
+class UnchainedAccountsPassword extends AccountsPassword {
+  async createUser(user, options) {
```
pozylon, line 70: Can you explain why you overwrite createUser?

pozylon:
Hey harry, It's really a good base but I think we really have to dig into this again a little deeper. After browsing through the reference of accountsjs I have the feeling that we can solve some things more elegant and more in line with the concepts of accountsjs.

packages/core-accountsjs/core-accountsjs.js
harryadel: Because the original implementation doesn't allow use to pass the options argument which we later need in our hooks.
https://github.com/accounts-js/accounts/blob/master/packages/password/src/accounts-password.ts#L593

tests/helpers.js
harryadel: Done.

packages/api/resolvers/mutations/accounts/createUser.js
harryadel: >If context is not submitted to createUser, it will be missing in any hooks regarding the createUser event, no?

Correct. The context wouldn't be available in the hooks. 

>Still the case? Also for changePassword and I guess you should pass it with your evaluateContext function
I have a little epiphany which is why I was reluctant to implement this request for two reasons:
- The implementation works already as is without the need for context; when we need it we can. And surely, there's a very good chance that we do  but still.
- Sometimes to add the context, this'd require us to extend accountsjs methods as was the case for `createUser` https://github.com/unchainedshop/unchained/pull/230#discussion_r497419058
and shall be for `changePassword` if we'd want to pass the context.
https://github.com/accounts-js/accounts/blob/master/packages/password/src/accounts-password.ts#L408

packages/api/resolvers/mutations/accounts/loginAsGuest.js
harryadel: This totally depends on whether the context was needed in later operations (which we did in `LoginSuccess` hook) and if method allows for passing the context.

https://github.com/accounts-js/accounts/blob/master/packages/server/src/accounts-server.ts#L176

Though the second condition of the possibility is much weaker since if there's a use case for a hook then I'd extend the methods.

packages/core-accountsjs/core-accountsjs.js
harryadel: The defer is part of the original implementation. See https://github.com/accounts-js/accounts/blob/master/packages/password/src/accounts-password.ts#L648

I tried testing things out locally and they work fine but still unsure about the change so I decided to ask for help [here](https://spectrum.chat/accounts-js/general/usage-of-lodash-defer-in-password-package~064218c6-707d-4576-9e4e-85dc16bcd386).

packages/core-accountsjs/core-accountsjs.js
pozylon: Hm, I see...

As it's class based, maybe you could you assign the context to the AccountServer in a way that it's available on "this", so you could go like this.unchainedContext in these methods. Maybe that could work so we could rely on the accountsjs code?

packages/platform/setup-accounts.js
pozylon: I've just seen the sanitizer function is only used for e-mail verifications but the CreateUserSuccess event is still likely the wrong place and facebook, google code obsolete

pozylon: (you can break login with facebook or login with google stuff)

packages/core-accountsjs/core-accountsjs.js
pozylon: Another thought: skipping e-mail verifications when a guest user is created could propably be skipped at a later time? maybe in server.prepareEmail or server.sendMail (we will have to touch these anyway later in the process)

packages/core-accountsjs/core-accountsjs.js
harryadel: On this branch, no. This was merely a basic implementation as it'd come in handy later. See https://github.com/unchainedshop/unchained/pull/241/files#diff-d1c5cccd98a611f535b5ad5f7c7786d9R130. It's not used anywhere in this branch but used to build upon in #241.

packages/core-accountsjs/core-accountsjs.js
harryadel: Good point on passing `this.unchainedContext`. I've had a closer look there's [sendVerificationEmail](https://github.com/accounts-js/accounts/blob/master/packages/password/src/accounts-password.ts#L650) option but the only downside to it is that we can't fine tune it per request. Yet, if it's ok with you we can actually work with it.

pozylon:
I had to merge master into accounts-js, resolve some conflicts because of some changes in the api package of me

pozylon:
I've merged this branch into "new-accounts" branch (not master) and will base the logout changes on that branch

PR #242, opened 2020-09-29
User search dropdown

harryadel:
Adding the front end part for https://github.com/unchainedshop/unchained/issues/226. The `SearchDropdown` component is getting reused more and more, so I decided to move the graphql search queries outside and have them set as props to allow for more general use cases.


pozylon:
Very good decision, everything feels much better now + you fixed an issue with assortment -> product links still using the inperformant query.

But there is one important thing that does not work atm. The user dropdown does not use your new queryString field on Query.users ;)

pozylon (changes requested):
Adjust user search query to actually search for users with the new queryString

PR #244, opened 2020-10-02
New Accounts Umbrella

pozylon:
This is the umbrella branch containing all accounts changes conducted by @harryadel


packages/api/resolvers/mutations/accounts/verifyEmail.js
```diff
@@ -1,5 +1,19 @@
...
+import {
+  accountsServer,
+  accountsPassword,
+  dbManager,
+} from 'meteor/unchained:core-accountsjs';
 
 export default async function verifyEmail(root, { token }, context) {
-  return callMethod(context, 'verifyEmail', token);
+  const user = await dbManager.findUserByEmailVerificationToken(token);
+  await accountsPassword.verifyEmail(token);
```
pozylon, line 9: Could this possibly return false and what if it does?

pozylon:
Dude, this looks goooood. Will plan an extended hour of tests with the code, the functionality and audit it for security issues.

Please rebase and fix the linting stuff

packages/api/resolvers/mutations/accounts/verifyEmail.js
harryadel: Given a false token, it definitely would. TBH, I've thought about wrapping all the mutations of `packages/api/resolvers/mutations/accounts` with try/catch but was unsure of what type of errors I should be throwing. Like vague ones tailored to each request i.e 'Change password request failed` or what. What do you think?

pozylon (commented):
- [ ] addEmail does not send a verification email and also does not immediately return the unverified email, maybe async/await? manually resending the verification e-mail works seamless

pozylon:
Amazing, the token stays, no fresh login needed !

pozylon:
- [ ] Please remove the german text in the verification mails and convert to harry english :)

pozylon:
I think we should write tests to make sure that the mailing with accounts works as expected, user auth api's and these kind of things are really important and very often used

pozylon:
@schmidsi @macrozone Can you guys please scan the branch and look for security issues that could have come up with the new accounts system?

pozylon:
@harryadel I started testing and unfortunately, the first thing I tried was not working so I stopped.

1. Case "Enroll Account No Password" -> Correct E-Mail generated -> Trying to set password -> Error: Reset password link expired

Please in-depth test every case you can think of again, fix that and also the codeship build (check is red atm) and re-request review.

PR #248, opened 2020-10-13
Fix broken tests

harryadel:
This PR entails fixing malfunctioning tests resulted from the previous accountsjs changes and also the rest of the conversion.


packages/core-users/db/helpers.js
```diff
@@ -106,38 +109,40 @@ Users.helpers({
...
       'user.updateEmail is deprecated, please use user.addEmail and user.removeEmail',
       { level: 'warn' }
     );
-    Accounts.addEmail(this._id, email, verified);
+    accountsPassword.addEmail(this._id, email, verified);
     (this.emails || [])
       .filter(({ address }) => address.toLowerCase() !== email.toLowerCase())
-      .forEach(({ address }) => Accounts.removeEmail(this._id, address));
+      .forEach(({ address }) =>
+        accountsPassword.removeEmail(this._id, address)
```
pozylon, line 142: Returns promise, shouldn't we have an await Promise.all here to wait until all emails are removed?

packages/core-users/db/helpers.js
```diff
@@ -76,9 +79,9 @@ Users.helpers({
       return profile.displayName;
     return emails && emails[0].address;
   },
-  setPassword(password, options) {
+  setPassword(password) {
     const newPassword = password || uuidv4().split('-').pop();
-    Accounts.setPassword(this._id, newPassword, options);
+    accountsPassword.setPassword(this._id, newPassword);
```
pozylon, line 84: Is this synchronous?

pozylon (commented):
Loogs good in general but I have the feeling that some of the methods that are changed to be async are not awaited in some calling functions, I'd double check those functions.

Also I can't find the place where you use "addWork" and reuse the existing messaging, but I guess this is something you'll tackle after we merge this PR?

PR #249, opened 2020-10-13
Refine new-accounts

harryadel:
This PR is intended to improve the changes made to new-accounts with proper testing and documentation.


packages/core-accountsjs/core-accountsjs.js
```diff
@@ -41,37 +41,44 @@ export const dbManager = new DatabaseManager({
...
+      });
+    }
   },
 };
 
 class UnchainedAccountsPassword extends AccountsPassword {}
 
-export const accountsPassword = new UnchainedAccountsPassword({});
+export const accountsPassword = new UnchainedAccountsPassword({
+  sendVerificationEmailAfterSignup: true,
```
pozylon, line 80: This has to be configurable from the outside, I think it was before by using "skipVerificationEmail" or something like that, else LGTM

pozylon (changes requested):

packages/core-accountsjs/core-accountsjs.js
pozylon: See an example on how to achieve "module" configurability:

https://docs.unchained.shop/config/assortments/

Technically:

https://github.com/unchainedshop/unchained/blob/master/packages/core-assortments/settings.js
-> settings provide a default value for everything

https://github.com/unchainedshop/unchained/blob/master/packages/core-assortments/assortments.js -> settings.load...

https://github.com/unchainedshop/unchained/blob/master/packages/core-assortments/db/helpers.js
-> importing settings

harryadel: You'd fine out that `accountsPassword` is accessed to set the variable internally, that's because it we tried to read the value off of settings like so `export const accountsPassword = new UnchainedAccountsPassword({
  sendVerificationEmailAfterSignup: settings.sendVerificationEmailAfterSignup,
});
` it'd fail to reevaluate after the configs are set. 

```
import { accountsPassword } from './accounts-password';

const sendVerificationEmail = true;

const settings = {
  load({ sendVerificationEmailAfterSignup = sendVerificationEmail } = {}) {
    accountsPassword.options.sendVerificationEmailAfterSignup = sendVerificationEmailAfterSignup;
  },
};
```

pozylon: Good solution in that case I guess :)

PR #250, opened 2020-10-20
New accounts mail fixes

harryadel:
This PR is intended for the email issues suggest by @pozylon.


packages/core-users/db/helpers.js
```diff
@@ -219,10 +207,9 @@ Users.enrollUser = async ({ password, email, displayName, address }) => {
     }
   );
   if (!params.password) {
+    const userEmail = Users.findOne({ _id: newUserId }).primaryEmail()?.address;
     // send an e-mail if password is not set allowing the user to set it
-    accountsPassword.sendEnrollmentEmail(
-      Users.findOne({ _id: newUserId }).primaryEmail()?.address
-    );
+    accountsPassword.sendEnrollmentEmail(userEmail);
```
pozylon, line 212: This can also be removed

packages/core-accountsjs/accounts-server.js
```diff
@@ -9,6 +9,15 @@ const accountsServerOptions = {
   siteUrl: process.env.ROOT_URL,
   prepareMail: (to, token, user, pathFragment, emailTemplate, from) => {
     if (token && pathFragment) {
+      // we're not supposed to be sending verifications to guests
+      // nor send verification emails to just **passwordless** enrolled users
+      if (
+        (user.guest || !user?.services?.password?.bcrypt) &&
```
pozylon, line 15: If (pathFragment === "verify-email" && user.isInitialPassword()) return {
        recipientEmail: to,
        action: 'enrollAccount',
        userId: user.id || user._id,
        token,
      };

packages/core-accountsjs/errors.js
```diff
@@ -0,0 +1,26 @@
+export default {
```
pozylon, line 1: remove

pozylon (changes requested):
Please check if isInitialPassword still works appropriately (set to true when enrollAccount and set to true when loginAsGuest), maybe you could change isInitialPassword to check for the bcrypt like you did in prepareEmail

packages/core-users/db/helpers.js
harryadel: The email-verification is the one triggered by default but enrollment email isn't so we need to keep it.

packages/core-accountsjs/errors.js
harryadel: Good catch!

packages/core-accountsjs/accounts-server.js
harryadel: This relates to the previous comment. We need to get in the way of verify-email which is fired automatically yet let enroll-accounts be sent freely without any interference.

packages/core-users/db/helpers.js
pozylon: Yeah exactly that's why I want to remap the verify-email to enrollAccount action in prepareEmail ;)

packages/core-accountsjs/accounts-server.js
pozylon: I'm not sure if you can use the current implementation of isInitialPassword here so please check carefully.

packages/core-users/db/helpers.js
pozylon: -> No we can remove it here

Issue #189, opened 2020-07-17
Improve Integration Test Surface


pozylon:
We have added a lot of checkboxes @ https://github.com/unchainedshop/unchained/issues/37. Now we also have a lot of tests but it's unclear how far we want to go with the integration tests. Also based on the decision about that test strategy, we will have to analyze the whole API and find out where we need to add more tests and more test cases.

**2 Examples of different "level of test depth"**

Just tests that addCartQuotation does not return an error:
https://github.com/unchainedshop/unchained/blob/master/tests/cart-quotations.test.js

Tests the whole payment process, valid and invalid payments, checks the db and the returning fields for feasability, also checks the underlying mongodb collection for certain data (https://github.com/unchainedshop/unchained/blob/master/tests/plugins-datatrans.test.js#L185):
https://github.com/unchainedshop/unchained/blob/master/tests/plugins-datatrans.test.js

I'm proposing the following approach:
- [x] Find out which mutations and queries are not tested at all and list those
- [x] Define general rules about what **integration** tests should test and what they should not test
- [ ] Apply the rules to all existing tests and create a github issue labeled with "good first issue, tests-needed" for every missing/insufficient test

@schmidsi @claudiocro @macrozone @Mikearaya Any ideas or feedback from your side?

Mikearaya:
This Are the queries and mutations that don't have any tests currently 

### Queries
- [ ] product
- [ ] products
- [ ] subscription
- [ ] subscriptions
- [ ] work


### Mutations
- [ ] activateSubscription
- [ ] addProductMedia
- [ ] heartbeat
- [ ] markPaymentCredentialsPreferred
- [ ] removePaymentCredentials
- [ ] removeWork
- [ ] setUsername
- [ ] terminateSubscription
- [ ] updateProductPlan
- [ ] updateSubscription

Mikearaya:
In my opinion testing, all scenarios would be great
- Testing for success when passed the expected data and also failure when passing invalid data.
- Test all possible scenarios where the endpoint should fail and also check if it's failing for the right reason, meaning not only checking if the `error` array contains something but for that specific error it should be returning.
- touch all conditions the request might pass `(if/else)`
- Finally, all endpoint that require special privileges  should be tested at least ones with a user that doesn't have that privilege

pozylon:
> Finally, all endpoint that require special privileges should be tested at least ones with a user that doesn't have that privilege

I like this very much.

PR #252, opened 2020-10-20
Integration tests #189

Mikearaya:
While testing I have noticed.

#### **Query.products**
 -  if `tags` and `slugs` are both specified as an argument then slugs will take precedence and only products that include the slugs will be returned and not tags.
- if a normal or non logged in user tries to get `draft` or `configurable` product by specifying `includeDrafts`: `true` then `NoPermissionError` is throws 

### **Query.subscriptions**
- only admin is getting subscriptions list

### **Query.subscription**
- :bug:  Found a bug here :bug: 
 property `isExpired` has a resolver with wrong logic. it tested `new Date` first which is always true then checks `referenceDate` that is never reached
```
now = new Date() || referenceDate
```
I've fixed it to instead
```
now =referenceDate || new Date() 
```
#### updated
this bug still exists so I've replaced it with the old one, I think there is something wrong with Date scalar conversion or something. whenever I try passing a reference date the result is not boolean but null. I was hoping you might have some clue. what do you think?

### **mutation.activateSubscription**
- returns throws a GERERIC **No suitable subscription plugin available for this plan configuration** when trying to activate a subscription that has a wrong product plan, should we make this error more specific?

### **mutation.updateSubscription**
- :bug:  return type was `Order` when it was actually returning `Subscription` :bug: 
so  I've fixed it accordingly
- I found some anomaly when passing the `plan` argument doesn't change the quantity of the product ???.
- also updating subscription as a normal user is possible but reading a single subscription is not, I think this contradicts. what do think about this?

### **mutation.heatbeat**
- it's not updating logs or activity of the user currently. Do you want to elaborate on the purpose of it, please?
- I just checked if it's returning the correct user information each time and if it returns an error for anonymous users calling it.

### **mutation.updateProductPlan**
- nothing happens when trying to update a product plan of a product that is not the type of `PLAN_PRODUCT`, shouldn't we inform that the product is not the right type to update it's plan and it should be `PLAN_PRODUCT` type or something? just a thought...

### **mutation.markPaymentCredentialsPreferred**
- admin can actually mark preferent payment credential of other users :thinking: 
- the API was not returning the updated PaymentCredential so I've updated that

### **mutation.removeWork**
- seems to remove a work no matter it's status


pozylon:
Amazing, will go through your discoveries step by step and get back to you soon!

pozylon:
Hey @Mikearaya can you split up the feedback into single issue we can discuss on? Just Query.subscriptions feedback is false positive, only admins should be able to get all subscriptions, that's fine

Mikearaya:
@pozylon ok, I'll

PR #253, opened 2020-10-21
Meteor roles migration

harryadel:
This PR is based on #244 as the user is intertwined with the roles and we're ought to make sure both work successfully. 

Our main goal was to remove underscore as much as possible, I also tried to trim some code yet this introduced some bugs so I decided to be ultra frugal with such removals and stick to our original goal of removing underscore.

We may need to delay the review of this PR until #244 is done.


harryadel:
Alright, so thanks to a rigorous review :wink: Lodash got further cut down, so we now only use clone module. I tried replacing it with spread operator but it led some tests to fail. I ain't offsetting my work here but would you please give it a go @pozylon?

pozylon:
How would you improve this package? Maybe we need some sort of unit tests for the logic here as it's very central for Unchained's security? if you rebase the new integration tests should cover a lot of permission surface tough

packages/roles/package.js
```diff
@@ -9,13 +9,16 @@ Package.describe({
...
+  });
+
   api.use([
     'unchained:core-users',
     'meteor-base',
     'check',
     'mongo',
     'ecmascript',
-    'underscore',
     'dburles:collection-helpers@1.1.0',
```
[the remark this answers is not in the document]
harryadel: The `check` package is used in many places in this packages. Since our main goal is to break free from Meteor, I can replace it with an NPM package instead.

packages/api/roles/loggedIn.js
```diff
@@ -81,9 +81,12 @@ export default (role, actions) => {
     return isOwnedOrder(null, { orderId }, { userId });
   };
 
-  const isOwnedProductReview = (root, { productReviewId }, { userId }) =>
-    ProductReviews.findReviewById(productReviewId).userId === userId;
-
+  const isOwnedProductReview = (root, middle, { userId }) => {
+    return (
```
pozylon, line 85: This results in 2 db calls instead of just 1, use optional chaining

packages/roles/package.js
```diff
@@ -9,22 +9,16 @@ Package.describe({
...
-    'check',
-    'mongo',
-    'ecmascript',
-    'underscore',
-    'dburles:collection-helpers@1.1.0',
-    'unchained:core-users@0.55.2',
-  ]);
+  Npm.depends({
+    'lodash.clone': '4.5.0',
+    '@share911/meteor-check': '1.0.8',
```
pozylon, line 14: I think you don't need this dep?

tests/jest-meteor-stubs/core-users.js
```diff
@@ -0,0 +1,26 @@
+const sinon = require('sinon');
```
pozylon, line 1: Nice

pozylon (commented):
Linter throws warnings about unnamed functions, please make sure to configure eslint properly in your IDE so it loads the eslint config file

pozylon:
Please move the role tests into the role package

pozylon:
I think we can drop check completely as Apollo guards us against

tests/jest-meteor-stubs/core-users.js
pozylon: Oh, now i See it's require style, please use import

harryadel:
>I think we can drop check completely as Apollo guards us against wrong serialization of parameters through the type system

You have a valid point for sure but I'm somewhat intimidated to implement it :sweat_smile:

pozylon:
> > I think we can drop check completely as Apollo guards us against wrong serialization of parameters through the type system
> 
> You have a valid point for sure but I'm somewhat intimidated to implement it 😅

@schmidsi Do you can think of any security related issue when not asserting the user input for the roles functions with a "check" like feature?

harryadel:
@pozylon Shoot first, ask later :smile: It's already removed.

harryadel:
@pozylon On a second note, using `check` would help us in cases when we don't pass data from Apollo and rather invoke the functions within our code, it'd help us catch bugs prematurely (i.e passing an array instead of a string). What do you think? Your call, keep it or remove it.

pozylon:
@harryadel We could help with that by converting this package to typescript or add typescript definitions. Which would make a lot of sense I guess

harryadel:
@pozylon I'm a typescript n00b but I'll do my best now and we can improve it together. :muscle:

pozylon:
Haha me too

harryadel:
@pozylon Will you please help me out with Codeship? I'm thinking because of the new step, do I need to modify any other CodeShip related files?  
```
- name: Packages Unit Tests
      service: ci
      command: meteor npm run test:packages:ci  
```

Pascal

examples/minimal/.meteor/versions
```diff
@@ -25,16 +25,18 @@ email@2.0.0
...
 littledata:synced-cron@1.5.1
-livedata@1.0.18
 logging@1.1.20
 meteor@1.9.3
+<<<<<<< HEAD
+minimongo@1.6.0
+=======
 meteor-base@1.4.0
 minimongo@1.6.1
+>>>>>>> 7581b3f2f1dd6869266d7158f7ba4218d0991897
```
pozylon, line 39: It can't possibly work at all with these kind of merge comments in the versions file, please do QA, look at the diffs

pozylon (changes requested):

harryadel:
The build stopped failing after removing the packages test step, so I restructured the commands so that both integration and unit tests are invoked in one go.

```
"test:ci": "npm run test:integration && npm run test:packages",
"test:integration": "jest --ci --runInBand --detectOpenHandles --forceExit tests/*",
"test:packages": "jest --ci --runInBand --detectOpenHandles --forceExit packages/*",
```

pozylon:
@schmidsi I need you on this because of the typescript typings, can you look at this and give @harryadel feedback? It should be possible to extend the available roles and permissions in user-land code.

schmidsi:
> @schmidsi I need you on this because of the typescript typings, can you look at this and give @harryadel feedback? It should be possible to extend the available roles and permissions in user-land code.

Yes, it is possible with [generics](https://www.typescriptlang.org/docs/handbook/generics.html), although generics is one of the more complicated concepts of types. I'm happy to write a test or example in the "minimal" (which is actually maximal :D) project, but we should first merge the typescript config from the `typescript`/`bookmark-refactoring` branch.

@harryadel Can you please merge the typescript config in and give me a heads up?

schmidsi:
> > > I think we can drop check completely as Apollo guards us against wrong serialization of parameters through the type system
> > 
> > 
> > You have a valid point for sure but I'm somewhat intimidated to implement it 😅
> 
> @schmidsi Do you can think of any security related issue when not asserting the user input for the roles functions with a "check" like feature?

User input checks should happen at resolver level. Having more checks deeper down is unnecessary complex in my opinion. Types will already help a lot.

pozylon:
@harryadel Can you please rebase this on the newest master? make sure the tsconfig of master is used

About `check`: @schmidsi I agree that user-input should be checked at resolver level but check is runtime, types is build-time so types will not help guarding our code from security breaches

pozylon (changes requested):
build fails

schmidsi:
> @harryadel Can you please rebase this on the newest master? make sure the tsconfig of master is used
> 
> About `check`: @schmidsi I agree that user-input should be checked at resolver level but check is runtime, types is build-time so types will not help guarding our code from security breaches

I agree: A run type check needs to happen at resolver level, types do not help there.

pozylon (changes requested):
@schmidsi It's updated, @harryadel please look at the changelog.md a version number is gone but a block is duplicate now, else looks good for me.

Also I looked closer and there is still some Meteor-ish stuff that you could remove:

For ex. line 345 in roles.js (Mongo.Collection.prototype.attachRoles). It uses this.allow and this.deny on the collections which only makes sense together with DDP and Unchained does not support DDP so you could propably strip

Mongo.Collection.prototype.attachRoles
Roles.Role.prototype.helper
Roles.registerHelper

And code only used by these functions but I'm not sure.

harryadel:
@pozylon I'm not capable of finding the lines you mentioned. Are you sure you revised the latest changes? `roles.js` no longer exists; only `roles.ts` which is ~200 lines.

pozylon:
I really like how modern this package feels now. It's a major step forward and super clean

PR #255, opened 2020-10-24
Fix toggle error

harryadel:
On fresh new installation the `updateHasMore` would be undefined yet it's triggered so we perform a simple guard to ensure that it exists.


pozylon:
nice!

PR #257, opened 2020-10-27
Enrollment fixes

harryadel:


packages/api/resolvers/mutations/accounts/index.js
```diff
@@ -17,5 +18,6 @@ export default {
   forgotPassword,
   resetPassword,
   resendVerificationEmail,
+  sendEnrollmentEmail,
```
pozylon, line 21: Hm, i know this is maybe a little off topic but shall we allow anonymous users to trigger verification and enrollment e-mails? at least the enrollment mail should only be triggerable by a user that can call "enrollUser" I think.

And for the resendVerificationEmail it does only make sense for the user that is logged in itself.

pozylon (approved):

PR #270, opened 2020-11-03
Meteor files migration

harryadel:
This PR started out by copying over the Meteor-Files package and removing any client side related code then the next step was to de-meteorize the code as much as possible and remove any Meteor dependencies but soon two problems arose: There're couple of associated Meteor packages like Webapp which is also used within Meteor-Cookies and unclear goals of what we would consider a successful migration. So, a new set of goals were defined thanks to @pozylon :

- Replace ostrio:files with a modern, minimal streaming capable file upload/download abstraction.
	- The implementation should only use raw mongodb-native as Dataloaders are created per request and wouldn't need work nicely with Meteor's mongo. In short, Avoid using globals like meteor/mongo.
	- We should rely on dependency injection style like what's been done in accountsjs implmentaion.

```
import MongoDBInterface from '@accounts/mongo';
import { MongoInternals } from 'meteor/mongo';
import { DatabaseManager } from '@accounts/database-manager';
import { idProvider, dateProvider } from './helpers';

const mongoStorage = new MongoDBInterface(
  MongoInternals.defaultRemoteCollectionDriver().mongo.db,
  {
    convertUserIdToMongoObjectId: false,
    convertSessionIdToMongoObjectId: false,
    idProvider,
    dateProvider,
  }
);
```
https://github.com/unchainedshop/unchained/blob/master/packages/core-accountsjs/db-manager.js
 
- Support S3 block storage
- Support media transformation
 	- Thumbnailing integration functionalities with thumbor for example
or movie downsampling with ffmpeg.

Additional theory:
https://martinfowler.com/books/refactoring.html
https://www.amazon.ca/Working-Effectively-Legacy-Michael-Feathers/dp/0131177052


harryadel:
The second and third goals are somewhat futuristic and pertaining our principle of "What we don't need now we don't need now" our immediate focus should be directed to replacing the currently used APIs while keeping in mind our future goals.

harryadel:
@pozylon While there're some tests to be added/fixed but architecture-wise this package is ready to be reviewed as we wrap things up for merging.

harryadel:
@pozylon 
Okidoki, so here're the latest news:

- I tried extending the unit tests suite especially the part about secured and unsecured collections but my attempts were proven unsuccessful because of how we mock `Webapp` (check `unchained/tests/jest-meteor-stubs/webapp.js`) which further discouraged me from trying to add any more unit tests. Maybe in the future where we're completely off of Meteor, unit testing would be more feasible for this package.

- `pretest` command was picking up `dataloader` and unable to resolve its respective node_modules types so at first I solved it like how I solved `reactioncommerce/random` and `file-type` by telling typescript to search into `api` package node_modules like so: 
```
"*": ["./packages/core-files/.npm/package/node_modules/*", "./packages/api/.npm/package/node_modules/*"],
```
But then this error popped up: 

```

> @unchainedshop/engine@0.61.0 typecheck /home/harry/Projects/unchained
> tsc

node_modules/@apollo/client/core/index.d.ts:14:10 - error TS2614: Module '"../../../../packages/api/.npm/package/node_modules/ts-invariant/lib/invariant"' has no exported member 'setVerbosity'. Did you mean to use 'import setVerbosity from "../../../../packages/api/.npm/package/node_modules/ts-invariant/lib/invariant"' instead?

14 import { setVerbosity } from "ts-invariant";
            ~~~~~~~~~~~~
```

so finally after much tinkering it worked with the current setup:

```
  "file-type":["./packages/core-files/.npm/package/node_modules/file-type/index.d.ts"],
      "@reactioncommerce/random": ["./packages/core-files/.npm/package/node_modules/@reactioncommerce/random"],
      "dataloader": ["./packages/api/.npm/package/node_modules/dataloader/index.d.ts"]
   ```
   
Now, I'm unsure if this is the best practice typescript-wise but it works for now. Maybe @schmidsi can shed some light.

- Remember the annoying tests I told you about in Basecamp? well that's why the build is failing. They're nothing but 4 simple tests but someone with a good insight to how the engine works especially the latest changes would solve them in an instant -- I already gave them a go but they were too much of a time sink. So, I'd appreciate if you @pozylon can handle those. 
But, aside from those aforementioned tests, I'd happy to resolve any of your comments.

All in all, the package could use more testing but due to the idiosyncrasies that involve using Meteor, Typescript and Jest it's a bit cumbersome right now and this package is long overdue. So, the best course of action to take it in a test-run on staging environment and fix any issues before merging.

Also, some side note for future maintainers, these are the collection that utilize files collection currently and their associated methods:

```
FilesCollection:
- OrderDocuments
  - insertWithRemoteURL
  - insertWithRemoteBuffer
- Media
  - insertWithRemoteBuffer
- QuotationDocuments
  - insertWithRemoteURL
  - insertWithRemoteBuffer
- Avatars
  - insertWithRemoteFile
  - insertWithRemoteBuffer
```

**Comment by pozylon** [comment]
@harryadel I have the same issues like you with typecheck and it does not work at all, I even tried installing the packages but didn't help. In the meantime I've removed the typecheck testings from CI to make the builds pass on Codeship but I'd like to completely remove that step from pretest. @schmidsi Maybe you have an idea how to do this right with typecheck

**Comment by pozylon** [comment]
@harryadel The most up-to-date version of master has just passed Codeship CI, so rebasing on that should fix your tests. Can you try doing that before we go down the rabbit hole with the tests?

**Comment by harryadel** [comment]
Master got merged in and the changes got kept so we can discuss them before attempting to modify anything.

Regarding the temporary directory changes, I think they're obsolete by now because we weren't doing anything them; it's how `ostrio:files` originally worked. So, in the new implementation of Meteor Files we're not storing the files on filesystem anymore instead we're directly passing the buffer into the GridFS without the need for intermediary process.

**Comment by pozylon** [comment]
Oh didn't know that it's dropping the temp storage. Yeah, I think you can remove that fileType stuff there.

**Comment by pozylon** [comment]
I really like that it's directly streaming to GridFS, I was just wondering if this has any performance, memory or file size implications

**Comment by harryadel** [comment]
> Oh didn't know that it's dropping the temp storage. Yeah, I think you can remove that fileType stuff there.
Glad we're on the same page. They're removed now! :+1: 

> I really like that it's directly streaming to GridFS, I was just wondering if this has any performance, memory or file size implications

I don't know about that part but it should be easy to handle such cases if they pose a problem in the future as write & load are nothing but simple functions now:

```
async write(buffer, opts: Options) {
    if (helpers.isFunction(opts)) {
      opts = {};
    }

    const fileId = opts.fileId || Random.id();
    const FSName = fileId;
    const fileName =
      opts.name || opts.fileName ? opts.name || opts.fileName : FSName;

    const { extension } = await getExtension(fileName, buffer);

    opts.type = await getMimeType(buffer);

    if (!helpers.isObject(opts.meta)) {
      opts.meta = {};
    }

    const size = opts.size || buffer.length;

    const result = dataToSchema({
      name: fileName,
      meta: opts.meta,
      type: opts.type,
      size,
      userId: opts.userId,
      collectionName: this._name,
      extension,
    });

    result._id = fileId;

    this.checkForSizeAndExtension({ size, extension });

    this.insert(result, async (err, _id) => {
      if (!err) {
        const fileRef = this.findOne(_id);
        await this.storeInGridFSBucket.call(this, fileRef, buffer);
      }
    });
    return result;
  }
  ```

harryadel:
I hope you can take a look at the failing tests soon so we can be done with this PR around the next week maximum, thanks!

pozylon:
Hey @harryadel I don't have time for that at the moment but the problem has to lie in jest.config.js, the test files, the packages used or any tests changed as the master currently builds just fine. I'd try to remove all the tweaks you have made like x-shop-country which is not needed for successful tests. (set env LANG=de COUNTRY=ch) also use the npm run test:watch which sets the verbose flag, maybe the engine crashes during the tests due to some files related code?

pozylon:
> > Oh didn't know that it's dropping the temp storage. Yeah, I think you can remove that fileType stuff there.
> 
> Glad we're on the same page. They're removed now! 👍
> 
> > I really like that it's directly streaming to GridFS, I was just wondering if this has any performance, memory or file size implications
> 
> I don't know about that part but it should be easy to handle such cases if they pose a problem in the future as write & load are nothing but simple functions now:
> 
> ```
> async write(buffer, opts: Options) {
>     if (helpers.isFunction(opts)) {
>       opts = {};
>     }
> 
>     const fileId = opts.fileId || Random.id();
>     const FSName = fileId;
>     const fileName =
>       opts.name || opts.fileName ? opts.name || opts.fileName : FSName;
> 
>     const { extension } = await getExtension(fileName, buffer);
> 
>     opts.type = await getMimeType(buffer);
> 
>     if (!helpers.isObject(opts.meta)) {
>       opts.meta = {};
>     }
> 
>     const size = opts.size || buffer.length;
> 
>     const result = dataToSchema({
>       name: fileName,
>       meta: opts.meta,
>       type: opts.type,
>       size,
>       userId: opts.userId,
>       collectionName: this._name,
>       extension,
>     });
> 
>     result._id = fileId;
> 
>     this.checkForSizeAndExtension({ size, extension });
> 
>     this.insert(result, async (err, _id) => {
>       if (!err) {
>         const fileRef = this.findOne(_id);
>         await this.storeInGridFSBucket.call(this, fileRef, buffer);
>       }
>     });
>     return result;
>   }
> ```

It's important that you are using read and write streams when receiving uploads so that the uploaded file is not cached in memory.  http://mongodb.github.io/node-mongodb-native/3.6/api/GridFSBucket.html#openUploadStream. Also for the S3-uploading there is usually a stream feature. Assuming you use the minio js api: https://docs.min.io/docs/javascript-client-api-reference.html -> see the putObject takes a stream

harryadel:
> Hey @harryadel I don't have time for that at the moment but the problem has to lie in jest.config.js, the test files, the packages used or any tests changed as the master currently builds just fine. I'd try to remove all the tweaks you have made like x-shop-country which is not needed for successful tests. (set env LANG=de COUNTRY=ch) also use the npm run test:watch which sets the verbose flag, maybe the engine crashes during the tests due to some files related code?

Alright, will look into it as you recommended.



> > > Oh didn't know that it's dropping the temp storage. Yeah, I think you can remove that fileType stuff there.
> > 
> > 
> > Glad we're on the same page. They're removed now! +1
> > > I really like that it's directly streaming to GridFS, I was just wondering if this has any performance, memory or file size implications
> > 
> > 
> > I don't know about that part but it should be easy to handle such cases if they pose a problem in the future as write & load are nothing but simple functions now:
> > ```
> > async write(buffer, opts: Options) {
> >     if (helpers.isFunction(opts)) {
> >       opts = {};
> >     }
> > 
> >     const fileId = opts.fileId || Random.id();
> >     const FSName = fileId;
> >     const fileName =
> >       opts.name || opts.fileName ? opts.name || opts.fileName : FSName;
> > 
> >     const { extension } = await getExtension(fileName, buffer);
> > 
> >     opts.type = await getMimeType(buffer);
> > 
> >     if (!helpers.isObject(opts.meta)) {
> >       opts.meta = {};
> >     }
> > 
> >     const size = opts.size || buffer.length;
> > 
> >     const result = dataToSchema({
> >       name: fileName,
> >       meta: opts.meta,
> >       type: opts.type,
> >       size,
> >       userId: opts.userId,
> >       collectionName: this._name,
> >       extension,
> >     });
> > 
> >     result._id = fileId;
> > 
> >     this.checkForSizeAndExtension({ size, extension });
> > 
> >     this.insert(result, async (err, _id) => {
> >       if (!err) {
> >         const fileRef = this.findOne(_id);
> >         await this.storeInGridFSBucket.call(this, fileRef, buffer);
> >       }
> >     });
> >     return result;
> >   }
> > ```
> 
> It's important that you are using read and write streams when receiving uploads so that the uploaded file is not cached in memory. http://mongodb.github.io/node-mongodb-native/3.6/api/GridFSBucket.html#openUploadStream. Also for the S3-uploading there is usually a stream feature. Assuming you use the minio js api: [docs.min.io/docs/javascript-client-api-reference.html](https://docs.min.io/docs/javascript-client-api-reference.html) -> see the putObject takes a stream

We do make use of `openUploadStream` in `storeInGridFS` methods, specifically this line:
```
const uploadStream = gridFSBucket.openUploadStream(file.name, {
        contentType: file.type || 'binary/octet-stream',
        metadata,
      });
```
Or should this be further moved up the chain? I don't understand, please elaborate.

As for the S3 changes, would it be OK to not include them in this PR? The goal of this one is to only migrate to Meteor-less version of file uploads.

pozylon:
Adding to the streaming: 

The basic file uploading should rely on a gridfs write stream. A write function that takes a buffer object would then write the buffer into the write stream. 

But I think the function also needs support for streaming as File Uploads in Apollo allow this. So in that case, you could .pip() the read stream of apollo with the write stream of gridfs and uploading 10GB files will be possible

See https://www.apollographql.com/blog/file-uploads-with-apollo-server-2-0-5db2f3f60675/

pozylon:
S3 you can drop for the moment

harryadel:
Ever since I started this PR, no checks ever succeeded but now! :confetti_ball:

harryadel:
The root cause of tests failure was some helpers of QuotationsDocument which invokes `.each()` which turns out that Meteor Files exported a special type of [cursor](https://github.com/veliovgroup/Meteor-Files/blob/master/cursor.js#L257) instead of MongoDB normal one. It completely went over my head and I thought it was a problem with DB or other tests.

Another minor issue which was causing the import bulk tests to fail was invalid image links that return 404. Now, with the new implementation such problem is detected so it's important to use proper links for testing.

Mikearaya:
* std error logged when adding product media
```
(node:13577) DeprecationWarning: File upload property ‘stream’ is deprecated. Use ‘createReadStream()’ instead.
```
* removing product media only removed media link but kept media file collection record. i don't think this is the desired behavior so i removed record media collection too, but can revert back.

Mikearaya:
@pozylon other than the issues i pointed out above, everything seems to be working as expected, functionality wise

pozylon:
So as it seems to me this is also ready, right?

pozylon:
@Mikearaya ping

pozylon:
We need to find a solution for pdf's and video, Publicare depends on the ability to load pdf's from files.

pozylon:
@Mikearaya changes still needed here too

Mikearaya:
@harryadel I found a big performance issue with gridFSBucket, specifically when using it for video [this line](https://github.com/unchainedshop/unchained/blob/59a8ff2a13239f3c91b8beb3e468bf6542ac6e3b/packages/core-files/createFilesCollection/FilesCollection.ts#L471) it stalls the engine. do you have any reason for changing the default 255KB to 1 KB because it seems to fix the issue with video?
Thanks

harryadel:
@Mikearaya https://github.com/veliovgroup/Meteor-Files/blob/7f67a0885e120ad77085b191e3399add575a2d89/server.js#L128
The transition wasn't smooth so I didn't want to introduce more entropy by adding unsolicited changes but if you have your reasons by all means go for it.

Mikearaya:
@pozylon feel free to check this out. PDF and video support, and minor fixes added.

pozylon:
Wohoo, will do

packages/core-files/createFilesCollection/helpers.ts
```diff
@@ -0,0 +1,281 @@
...
+        return false;
+      }
+      obj = obj[path[i]];
+    }
+    return !!length;
+  },
+  now: Date.now,
+};
+
+export const accessDenied = (http, rc: number): boolean => {
```
pozylon, line 76: should be part of api package, see createBulkImportServer for a reference how to handle stuff at certain paths with permission checks and user context

packages/core-files/createFilesCollection/helpers.ts
```diff
@@ -0,0 +1,281 @@
...
+    return {
+      ext: fileExtension.ext,
+      extension: fileExtension.ext,
+      extensionWithDot: `.${fileExtension.ext}`,
+    };
+  }
+  return { ext: '', extension: '', extensionWithDot: '' };
+};
+
+export const responseHeaders = (responseCode, versionRef) => {
```
pozylon, line 127: should be part of api package, see createBulkImportServer for a reference how to handle stuff at certain paths with permission checks and user context

packages/core-files/createFilesCollection/helpers.ts
```diff
@@ -0,0 +1,281 @@
...
+  const fileExtension = await fileType.fromBuffer(buffer);
+  mime = fileExtension?.mime;
+
+  if (!mime || !helpers.isString(mime)) {
+    mime = 'application/octet-stream';
+  }
+  return mime;
+};
+
+const hashLoginToken = (stampedLoginToken: string): string => {
```
pozylon, line 200: This functions are part of accounts and duplicate, remove

packages/core-files/createFilesCollection/helpers.ts
```diff
@@ -0,0 +1,281 @@
...
+  const hash = crypto.createHash('sha256');
+  hash.update(stampedLoginToken);
+  const hashedToken = hash.digest('base64');
+
+  return hashedToken;
+};
+
+const getTokenLifetimeMs = () => {
```
pozylon, line 208: This functions are part of accounts and duplicate, remove

packages/core-files/createFilesCollection/helpers.ts
```diff
@@ -0,0 +1,281 @@
...
+  return (
+    (LOGIN_UNEXPIRING_TOKEN_DAYS || DEFAULT_LOGIN_EXPIRATION_DAYS) *
+    24 *
+    60 *
+    60 *
+    1000
+  );
+};
+
+const tokenExpiration = (when: Date): Date => {
```
pozylon, line 220: This functions are part of accounts and duplicate, remove

packages/core-files/createFilesCollection/index.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import FilesCollection from './FilesCollection';
+
+const createDefaultSettings = () => ({
+  maxSize: 10485760,
+  extensionRegex: null,
+});
+
+export default (collectionName, customSettings = {}) => {
+  const mergedSettings = {
```
pozylon, line 11: Where are all the other settings going despite maxSize and extensionRegex? ^^

packages/core-products/db/product-media/helpers.js
```diff
@@ -8,6 +8,8 @@ ProductMedia.findProductMedia = ({ productMediaId }) => {
 };
 
 ProductMedia.removeProductMedia = ({ productMediaId }) => {
+  const productMedia = ProductMedia.findProductMedia({ productMediaId });
+  Media.removeFiles({ _id: productMedia.mediaId });
```
pozylon, line 12: Not good, removing a product media does not automatically mean to remove the media, media's can be re-used

pozylon:
After the tests passed and bugs fixed I finally found the time to look at the structure and I think there is a long way to go mainly because of separation of concern and duplication of code:
- HTTP based stuff has to go to API package so that the same user and permission check functions apply to file downloads like for graphql and other stuff and ACL works.
- Auth code has to be removed completely
- We need a clear abstraction layer that allows us to add s3 *streamable* up- and download. Currently files are stored in temp folders and then passed along to gridfs hardcoded. Upload and Download streams can be passed to third party libraries and for example GridFS directly so we don't even have to store a single byte temporary.

@Mikearaya I'm not sure if we should continue to invest in this PR, we spent a lot of time on it but I think we just have to rethink it and build it from the start and take over the learnings we made...

PR #271, opened 2020-11-19
Multiple currency support

Mikearaya:
I've added the currency exchange from `EUR` to currency `X` where currency `x`  is a currency supported in European Central Bank exchange, also a function that will convert from any `base` currency to any `target` currency using coinbase API. this is mainly for the purpose of changing `fiat` to `crypto` or `crypto` to `fiat` but can be used for fiat to fiat if the user is comfortable using coinbase for exchange. 
we are currently using it for a plugin that converts any fiat to the 10 cryptocurrencies we are currently supporting namely  `BTC`, `ETH`, `XRP`, `USDT`, `BCH`, `BSV`, `LTC`, `EOS`, `BNB`, `XTZ`. I've selected this crypto based on their current market capitalization https://coinrivet.com/top-10-cryptocurrencies-by-market-capitalisation/

In addition, I've only enabled fiat to crypto exchange plugin because activating EUR to fiat plugin assumes that users are using EUR as their default currency which might not be the case. so instead it's all set up and ready for use with only an import when someone want to use this plugin.


packages/core-pricing/plugins/product-price-fiat-crypto-exchange.js
```diff
@@ -0,0 +1,46 @@
...
+    return AVAILABLE_CURRENCIES.crypto.indexOf(currency.toUpperCase()) !== -1;
+  }
+
+  async calculate() {
+    const { product, country, quantity, currency } = this.context;
+    const defaultCurrency = Countries.resolveDefaultCurrencyCode({
+      isoCode: country,
+    });
+    const productPrice = product.price({ country, currency: defaultCurrency });
+    if (!productPrice || !productPrice?.amount) return null;
```
pozylon, line 29: See ECB plugin, same comments apply

pozylon (changes requested):
Very good job, yet we can optimize this quite a bit so it feels good in all cases.

pozylon:
As pricing plugins get chained and hand over results it's very important that you don't return null. If you return null in a pricing plugin it means to cut the chain, which might not be desired, assuming you have these plugins activated and you want the following:

Pricing calculation for price in BTC:
1. Catalog Price (sort 0) -> results in no price because catalog price not there
2. ECB Plugin -> results in no price because EUR price not there
3. Coinbase Plugin -> results in a price because default currency CHF and Coinbase can do CHF -> BTC
4. Round to 50
5. Calculate VAT for Switzerland

packages/core-pricing/plugins/product-price-EUR-fiat-exchange.js
```diff
@@ -0,0 +1,42 @@
...
+  static orderIndex = 1;
+
+  static isActivatedFor({ currency }) {
+    return AVAILABLE_CURRENCIES.fiat.indexOf(currency) !== -1;
+  }
+
+  async calculate() {
+    const { product, country, quantity, currency } = this.context;
+    const EURprice = product.price({ country, currency: 'EUR' });
+    if (!EURprice || !EURprice?.amount) return null;
```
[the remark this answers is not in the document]
Mikearaya: i think 2 is already done with the line ` if (!EURprice || !EURprice?.amount) return null;` which will be change now based on remark 1 to  `if (!EURprice || !EURprice?.amount) super.calculate();`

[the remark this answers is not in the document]
pozylon: Yeah but you should check this.calculation too

Mikearaya:
@pozylon I agree it would be nice if round to the next 50 is a plugin on its on so anyone can use it when needed, but an issue comes once added and the rounding is needed only for certain currency type/s but not on others. eg. users might not want to round to the next 50 when the currency is in crypto because there is a significant difference when doing that but they may want to round to the next 50 on fiat currency.
so unless we disable rounding when currency is crypto-based explicitly there is no other way to disable this once the plugin is enabled. what is your thought on this? 
FYI I've extracted the logic into its own plugin already but I thought of this issue

pozylon:
That is a very good point, maybe we could configure the rounding plugin? So like:

```
import RoundPlugin from 'meteor/unchained:core-pricing/plugins/round';

RoundPlugin.configure({
  "BTC": { "precision": 5000, "mercantile": false },
}) // only btc rounded to 5000 sats
```

or something like this, and all non

Mikearaya:
for the product price rounding plugin configurable approach. now can do it like this...
it's just really a draft to make sure if we are on the same page.
```
import { ProductPricingDirector } from 'meteor/unchained:core-pricing';
import ProductPriceRound from './product-price-round';

ProductPriceRound.configure({ currency: 'USD', precision: 100 });
ProductPriceRound.configure({ currency: 'default', precision: 50 });

ProductPricingDirector.registerAdapter(ProductPriceRound);
```

packages/core-pricing/plugins/product-price-coinbase-exchange.js
```diff
@@ -0,0 +1,71 @@
...
+  );
+  return data?.rates?.[target];
+};
+
+class ProductPriceCoinbaseExchange extends ProductPricingAdapter {
+  static key = 'shop.unchained.pricing.price-coinbase-exchange';
+
+  static version = '1.0';
+
+  static label = 'Convert EUR to crypto with curreny exchange rate';
```
pozylon, line 38: This plugin is not restricted to EUR isn't it? It's more like "Convert Default Currency to Crypto...."

packages/core-pricing/plugins/product-price-coinbase-exchange.js
```diff
@@ -0,0 +1,71 @@
...
+  'BSV',
+  'LTC',
+  'EOS',
+  'BNB',
+  'XTZ',
+];
+const cache = new Cache(CACHE_PERIOD);
+
+const getFiatexchangeRateForCrypto = async (base, target) => {
+  const { data } = await cache.get(`${base}-${target}`, () =>
```
pozylon, line 25: ECB rates can be cached for 24h I thought, maybe define here specifically rather than the default 60s

pozylon (approved):
Very good! I only have found some very small things. I think we should provide some documentation about the available pricing plugins at some point.

Mikearaya:
we definitely need to provide documentation for the pricing, I'll add it under configuration and i think the rounding plugin will be good for example.

PR #273, opened 2020-11-25
Type assertions

Mikearaya:
* `mutation.createProductVariation`, `mutation.addProductAssignment`, `mutation.removeProductAssignment`, `mutation.createProductBundleItem` & `mutation.removeBundleItem` have business logic inside resolver functions. **i thinking moving this logic to the package is better design/architecture wise.**
* There is no CARD payment adapter currently supported so unfortunately no test is made for `mutation.mutation.updateOrderPaymentCard` but the test made for the others `mutation.mutation.updateOrderPaymentGeneric` & `mutation.mutation.updateOrderPaymentInvoice` apply the same logic so it's bullet proof.


package.json
```diff
@@ -21,7 +21,7 @@
     "lint": "eslint packages tests --cache --fix --ext=jsx --ext=js",
     "pretest": "npm run lint --silent",
     "lint:ci": "eslint packages tests --ext=jsx --ext=js",
-    "test": "jest --runInBand tests/*",
+    "test": "jest --runInBand",
```
pozylon, line 24: Why did you change that? It looks like now there is docs tests included in the test that fail

packages/api/resolvers/mutations/createProductBundleItem.js
```diff
@@ -1,16 +1,29 @@
...
+
   if (!productId) throw new InvalidIdError({ productId });
   if (!item.productId)
     throw new InvalidIdError({ bundleItemId: item.productId });
   const product = Products.findOne(productId);
-  if (!product) throw new ProductNotFoundError({ productId });
+
   const bundleProduct = Products.findOne(item.productId);
   if (!bundleProduct) throw new ProductNotFoundError({ productId });
+  if (!product) throw new ProductNotFoundError({ productId });
```
pozylon, line 20: The productId provided to the error here is not the item's productId. And also it looks like the bundleProduct is actually the item where as the product is the bundleProduct

pozylon: Maybe optimize the naming here along the way

packages/api/resolvers/mutations/createProductVariation.js
```diff
@@ -1,16 +1,32 @@
...
 ) {
   log(`mutation createProductVariation ${productId}`, { userId });
+
   if (!productId) throw new InvalidIdError({ productId });
   const product = Products.findOne({ _id: productId });
   if (!product) throw new ProductNotFoundError({ productId });
+  if (product.type !== ProductTypes.ConfigurableProduct)
+    throw new ProductWrongTypeError({
+      productId,
+      recieved: product.type,
```
pozylon, line 26: recieved -> received

packages/api/resolvers/mutations/removeBundleItem.js
```diff
@@ -1,12 +1,22 @@
...
 
 export default function removeBundleItem(root, { productId, index }) {
   log(`mutation removeBundleItem ${productId}`, { index });
   if (!productId) throw new InvalidIdError({ productId });
   const product = Products.findOne(productId);
   if (!product) throw new ProductNotFoundError({ productId });
+  if (product.type !== ProductTypes.BundleProduct)
+    throw new ProductWrongTypeError({
+      productId,
+      recieved: product.type,
```
pozylon, line 17: Same here, typo

packages/api/resolvers/mutations/createProductBundleItem.js
```diff
@@ -1,16 +1,29 @@
...
+  if (bundleProduct.type !== ProductTypes.BundleProduct)
+    throw new ProductWrongTypeError({
+      productId,
+      recieved: bundleProduct.type,
```
pozylon, line 24: "recieved"

packages/api/resolvers/mutations/addProductAssignment.js
```diff
@@ -11,11 +16,17 @@ export default function addProductAssignment(
 
   if (!proxyId) throw new InvalidIdError({ proxyId });
   if (!productId) throw new InvalidIdError({ productId });
+  const product = Products.findOne({ _id: productId });
+  if (!product) throw new ProductNotFoundError({ productId });
+  if (product.type !== ProductTypes.ConfigurableProduct)
+    throw new ProductWrongTypeError({
+      productId,
+      recieved: product.type,
```
pozylon, line 24: "recieved"

packages/api/resolvers/mutations/removeProductAssignment.js
```diff
@@ -1,16 +1,27 @@
...
 ) {
   log(`mutation removeProductAssignment ${proxyId}`, { userId });
+
   if (!proxyId) throw new InvalidIdError({ proxyId });
   const product = Products.findOne({ _id: proxyId });
   if (!product) throw new ProductNotFoundError({ proxyId });
+  if (product.type !== ProductTypes.ConfigurableProduct)
+    throw new ProductWrongTypeError({
+      productId: proxyId,
+      recieved: product.type,
```
pozylon, line 22: same

packages/api/resolvers/mutations/updateOrderDelivery.js
```diff
@@ -1,15 +1,35 @@
...
+  updateOrderDeliveryShipping: DeliveryProviderType.SHIPPING,
+  updateOrderDeliveryPickUp: DeliveryProviderType.PICKUP,
+};
 
 export default function updateOrderDelivery(
   root,
   { orderDeliveryId, ...context },
-  { userId }
+  { userId },
+  { fieldName }
```
pozylon, line 19: Good idea but I think we should solve that differently, having 2 explicit resolver functions instead of using fieldName and a custom map. Like this it's a little bit complex just for DRY's sake.

packages/api/resolvers/mutations/updateOrderPayment.js
```diff
@@ -1,16 +1,44 @@
...
-import { OrderPaymentNotFoundError, InvalidIdError } from '../../errors';
+import { PaymentProviderType } from 'meteor/unchained:core-payment';
+import {
+  OrderPaymentNotFoundError,
+  InvalidIdError,
+  OrderPaymentTypeError,
+} from '../../errors';
+
+const PAYMENT_UPDATE_ENDPOINT = {
+  updateOrderPaymentCard: PaymentProviderType.CARD,
```
pozylon, line 11: Same here, prefer composition

packages/api/resolvers/mutations/updateProductPlan.js
```diff
@@ -11,5 +15,10 @@ export default function updateProductPlan(
   if (!productId) throw new InvalidIdError({ productId });
   const productObject = Products.updateProduct({ productId, plan });
   if (!productObject) throw new ProductNotFoundError({ productId });
+  if (productObject?.type !== ProductTypes.PlanProduct)
+    throw new ProductWrongStatusError({
+      received: productObject?.type,
```
pozylon, line 20: here it's correct :)

packages/api/resolvers/mutations/updateProductWarehousing.js
```diff
@@ -1,15 +1,28 @@
...
   if (!productId) throw new InvalidIdError({ productId });
-  const productObject = Products.updateProduct({ productId, warehousing });
-  if (!productObject) throw new ProductNotFoundError({ productId });
-  return productObject;
+  const product = Products.findOne({ _id: productId });
+  if (!product) throw new ProductNotFoundError({ productId });
+  if (product?.type !== ProductTypes.SimpleProduct)
+    throw new ProductWrongTypeError({
+      productId,
+      recieved: product.type,
```
pozylon, line 23: again, "recieved"

pozylon:
I totally agree with you moving the business logic with the updates like you suggested to the packages!

package.json
Mikearaya: I removed it because I wanted to run a single file of test with `npm run test -- file-name` and couldn't do that with the previous setup. all tests seem to be working when running it by `npm run test` but I'll replace it if you have encountered an issue.

packages/api/roles/admin.js
```diff
@@ -31,8 +31,11 @@ export default (role, actions) => {
   role.allow(actions.updateOrder, () => true);
   role.allow(actions.updateOrderDiscount, () => true);
   role.allow(actions.updateOrderItem, () => true);
-  role.allow(actions.updateOrderDelivery, () => true);
```
pozylon, line 34: While I think it did make sense to duplicate the resolvers so we have a file per query/mutation this is not necessarily true for the actions. I can't think of any case where you want to allow a user to update the context of a pickup delivery but disallow updating the context of a shipping delivery

packages/api/resolvers/mutations/addProductAssignment.js
```diff
@@ -11,11 +16,17 @@ export default function addProductAssignment(
 
   if (!proxyId) throw new InvalidIdError({ proxyId });
   if (!productId) throw new InvalidIdError({ productId });
+  const product = Products.findOne({ _id: productId });
+  if (!product) throw new ProductNotFoundError({ productId });
+  if (product.type !== ProductTypes.ConfigurableProduct)
```
pozylon, line 21: I think this is wrong, proxyId product should be asserted as a configurable product, maybe use childProduct as variable name instead of product to not have us fall into this again.

packages/api/resolvers/mutations/createProductBundleItem.js
```diff
@@ -1,16 +1,30 @@
...
     throw new InvalidIdError({ bundleItemId: item.productId });
   const product = Products.findOne(productId);
+
+  const bundleItem = Products.findOne(item.productId);
+  if (!bundleItem)
+    throw new ProductNotFoundError({ productId: item.productId });
   if (!product) throw new ProductNotFoundError({ productId });
-  const bundleProduct = Products.findOne(item.productId);
-  if (!bundleProduct) throw new ProductNotFoundError({ productId });
+  if (bundleItem.type !== ProductTypes.BundleProduct)
```
pozylon, line 22: Also this, I think productId product has to be a BundleProduct where as the bundleItem is part of the bundle therefore can be anything.

pozylon (changes requested):

pozylon:
LGTM

PR #278, opened 2020-12-08
Allow updating quantity and product of a subscription

### LINKED ISSUE #264: mutation.updateSubscription doesn't update plans quantity
I found some anomaly, when passing the **plan** argument doesn't change the quantity of the product

**To Reproduce**
call mutation.updateSubscription with a plan argument

**Expected behavior**
when passed plan as an argument that has a different quantity than before the subscription should be updated accordingly.

pozylon:
This change will allow to change very basic subscription details like the actual product that defines the charging period and the quantity of the product, it fixes #264.

Currently, as long as a subscription is INITIAL, updatePlan will work accordingly but it fails when it's not INITIAL. Before merging this PR, we should add business logic to change an already active subscription's product.

What should happen in that case?

Active Subscription with Product A and 2 periods, one active:
1.2020 - 6.2020
6.2020 - 1.2021

Beeing at date 12.2020 we change the product to Product B which has another price and also is charged monthly so what happens is there will be another period added:

1.2020 - 6.2020
6.2020 - 1.2021 Product A
12.2020 - 1.2021 Product B

That's pretty fine but what if Product B is an upgrade to Product A? In that case we would like to see:
1.2020 - 6.2020
6.2020 - 15.12.2021 Product A
15.12.2020 - 15.1.2021 Product B

There is already an invoice for the former 5.2020 - 1.2021 Product A, so we have to pro-rata deduct and add this situation to the invoice like

15.12.2020 Invoice:
(01.01.2021 - 15.12.2021) Product A Upgrade Credit Note  = - XYZ USD
1 Month Product B = + XYZ USD
Total = Sum of Product A Credit Note and Product B Monthly Price

### DIFF
--- a/packages/api/resolvers/mutations/updateSubscription.js
+++ b/packages/api/resolvers/mutations/updateSubscription.js
@@ -41,7 +41,13 @@ export default async function updateSubscription(
     Subscriptions.updateDelivery({ delivery, subscriptionId });
   }
   if (plan) {
-    Subscriptions.updatePlan({ plan, subscriptionId });
+    if (subscription.status !== SubscriptionStatus.INITIAL) {
+      // If Subscription is not initial, forcefully add a new Period that OVERLAPS the existing periods
+      throw new Error(
+        'TODO: Unchained currently does not support order splitting for subscriptions, therefore updates to quantity, product and configuration of a subscription is forbidden for non initial subscriptions'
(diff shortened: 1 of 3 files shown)


pozylon:
I decided to remove the subscription plan splitting feature from the MR in order to merge

PR #279, opened 2020-12-15
Set Unchained colors

harryadel:
I copied the [site configurations](https://github.com/Semantic-Org/Semantic-UI-LESS/blob/e4395217c1b8b3227c7387284d12f2d9774d33c6/themes/default/globals/site.variables) from the default theme then modified the colors to match Unchained website but I don't know if this is enough.


pozylon:
good, can you change the logo also to match the current website? black/white

Issue #275, opened 2020-12-07
business logic should not be exposed at resolver level


Mikearaya:
currently some resolvers have business logic defined in them such as  `mutation.createProductVariation`, `mutation.addProductAssignment`, `mutation.removeProductAssignment`, `mutation.createProductBundleItem` & `mutation.removeBundleItem`. 
but this business logic should be denied inside a package that is responsible for that particular operation. 
Please identify each endpoint that has business logic defined in them and move the logic to the corresponding package.

Issue #246, opened 2020-10-06
do not throw error when resource is not found


macrozone:
**Describe the bug**

current unchained started to throw errors when a product or assortment is not found.

I have not a final opinion on that, but it seems to be an antipattern because it does not add any information to clients but rather makes stuff more complicated: https://stackoverflow.com/a/54641803

e.g. SSR will always throw errors when one request has an error, which is really annoying

**To Reproduce**

**Expected behavior**

resource not found should not throw an error but return null as before

macrozone:
there is also a case for having infos in the errors field: https://stackoverflow.com/a/57462949

so no conclusion for me. 

I think make errors more obvious is not a bad thing, but it should not break when errors were expected. E.g. user provided slugs etc.

schmidsi:
I think this is a bug. We refactored the internal errors to be more meaningful and this might have introduced this issue. As far as I understand the GraphQL spec: It should try to always respond and only throw if there is a fatal error on the server side. Like DB-connection interrupted.

But, I like the idea of having both a `data` and a `errors` fields in the response, as long as there was no fatal error on the server side. But this would possibly require changes on the front-ends, since a lot of developers have the following mind-pattern:

> If there is an error, there is no data and therefore we do not display anything.

It should be shifted to:

> If I have some data, display it.

macrozone:
> I think this is a bug. We refactored the internal errors to be more meaningful and this might have introduced this issue. As far as I understand the GraphQL spec: It should try to always respond and only throw if there is a fatal error on the server side. Like DB-connection interrupted.
> 
> But, I like the idea of having both a `data` and a `errors` fields in the response, as long as there was no fatal error on the server side. But this would possibly require changes on the front-ends, since a lot of developers have the following mind-pattern:
> 
> > If there is an error, there is no data and therefore we do not display anything.
> 
> It should be shifted to:
> 
> > If I have some data, display it.

indeed, there is also the errorPolicy option in apollo client to decide on what to do when there is an error.

Problem in our project is, that we have some branching logic based on wether an url path segment is an assortment or a product. this will now always trigger errors

Mikearaya:
this is not a bug, `assortment` or any resource requested by id is still returned as `null` when it's not found but now there is more clarity as to why it is null that give front-end more information to display an appropriate message based on the `error` returned from the server.

macrozone:
@Mikearaya maybe its intentional, but it will break many project. In particular it breaks SSR in many cases

[...]

Issue #115, opened 2019-11-05
[core] Strategy/Roadmap 1.0 (Draft)


pozylon:
After many discussins and various issues, hearing to the community and other developers, we have laid out a general roadmap on how we want to transform Unchained. 

**We want to get off the meteor train. Why?**
Using emerging technologies and architectures like typescript, jest, serverless configurations has been hard with a runtime that uses a custom bundler, a proprietary package manager and an always older than current node.js version. We are not convinced of the future of meteor and we think going a more agnostic way would open up the whole system to many more developers. Too many times we have faced meteor bias (does that scale? is that still maintainted?) that kept developers from using unchained in their next e-commerce projects.

# Vision: Become Node.js Framework agnostic

We should refactor unchained before we go 1.0 and this is a first try for an architectural draft. The current logic around helpers and direct module-to-module dependencies is not good enough in terms of separation of concerns and modularity. When doing this kind of big refactorings, it's key to do it in small steps and with a clear scope in min.

In Scope:
- Dataloader + TS + Context (Big GraphQL Refactor)
- Delete Fixtures / Factories
- Delete Migrations
- Migrate from Meteor Files
- Migrate from Accounts
- Migrate from all the other packages like SyncedCron (#213, #214, #215, #216)
- Move Code from Packages to one single NPM Package

Out Of Scope:
- MongoDB / ORM / Query Builders (we will use the nodejs mongodb native library and we will not try to make unchained db agnostic)
- Migrate from Simple Schema Validation (we will still use simple-schemas for the moment)
- Yarn Workspaces / Multi-package approach, ... (we will not change to some sophisticated workspace or package monster architecture)
- Deno :) (we will not use something else than Node.js for the moment)
- New Migration Tool (we will not evaluate and add a sophisticated migration tool and instead use a very lightweight custom one from a project)
- API Decoupling (We will not try to split up the GraphQL API definition to various modules and instead keep the GraphQL part in one single module like it is now)

## Milestones:

1. Dataloaders & Services
2. No external Meteor Packages (Accounts, Files, ...)
3. New NPM Package @unchainedshop/unchained

## Benefits:

1. Unit testable core modules -> no helpers, domain logic as functional code
2. Interdependencies / Modularity -> Extract glue code to it's own right (checkout flow for ex.), making the whole system more flexible and organized
3. Performance: Solve 1+n performance problem with Dataloaders for all DB access

# Architecture

## Layers

[...]

PR #281, opened 2020-12-21
business logic encapsulation #275

Mikearaya:
all the mutations resolver functions are now encapsulated. i don't know if we should take it to the query resolvers since they use methods on the collection itself and only do the logic relating to limit, offset, etc... what is your take on this?


pozylon:
Wow, I need some time to dig through this but looks awesome on first sight.

I think having the API package decoupled from Meteor's Mongo functions is great. It removes the need to unit test resolvers completely (the explicit errors are already tested through the integration tests) and along the way makes it clear where business logic is. Also it helps us isolate meteor/mongo dependency further.

So I think a good benchmark could be to go through all resolvers and check if no meteor/mongo functions are called directly anymore (find, update, upsert, ... https://docs.meteor.com/api/collections.html).

I think we have to go even further to make the packages unit testable and allow us to decouple the core-packages from each other and I have put my ideas and thinings into this comment here:

https://github.com/unchainedshop/unchained/issues/115#issuecomment-751680427

It's kind of like a culmination of your PR, my initial ideas and the ongoing refactorings

packages/api/resolvers/mutations/createProductBundleItem.js
```diff
@@ -25,14 +25,5 @@ export default function createProductBundleItem(root, { productId, item }) {
...
-    $set: {
-      updated: new Date(),
-    },
-    $push: {
-      bundleItems: item,
-    },
-  });
-
-  return Products.findOne(productId);
+  return product.createBundleItem({ item });
```
pozylon, line 28: I think this is more accurate like "addBundleItem"

pozylon: I mean the mutation is fine with "create" when you provide the item id and the product id but as a scoped function like "product.createBundleItem" it's a bit weird and it propably is better to have it more documenting what happens like:

"bundle.addItem()" or "product.addBundleItem()"...

packages/api/resolvers/mutations/setUserTags.js
```diff
@@ -11,14 +11,5 @@ export default function setUserTags(
...
-    { _id: normalizedUserId },
-    {
-      $set: {
-        updated: new Date(),
-        tags,
-      },
-    }
-  );
-  return Users.findOne({ _id: normalizedUserId });
+  return user.addTags({ tags });
```
pozylon, line 14: addTags means that the already added tags are preserved, this should be setTags because it actually overwrites all tags with the new array.

packages/api/resolvers/mutations/updateCart.js
```diff
@@ -15,21 +15,13 @@ export default async function updateCart(
...
-  if (paymentProviderId) {
-    order = order.setPaymentProvider({ paymentProviderId });
-  }
-  if (deliveryProviderId) {
-    order = order.setDeliveryProvider({ deliveryProviderId });
-  }
-  return order;
+  const order = await getCart({ orderId, user, countryContext });
+
+  return order.updateCart({
```
pozylon, line 20: I think it's better to revert moving this to an own updateCart helper because it's abstracted good enough already. It can be seen as the decision of the API designer to allow changing multiple aspects of an order with one updateCart mutation but it's not really core.

pozylon (changes requested):
I stopped reviewing because I think we have to change something quite fundamentally here 😬

As you can see in some packages I've already done some abstractions like for Products I've created a Products.createProduct function (https://github.com/unchainedshop/unchained/blob/780259b085339101468f83e8cbbfa4629fcef8fa/packages/core-products/db/products/helpers.js#L18) and also for other parts like Assortments there exist these kind of functions that alter the database.

Our plan currently is to get away from a collection helper approach to more immutable, functional design so we should avoid adding new collection helpers when the only purpose of the function is to fire some db updates because we will have to remove these anyway later in the process.

Right now you have added crud functions like updateLanguage, removeLanguage as helpers to the collection instance that is getting modified.

I think it's much better to have these on the prototype level like
Languages.updateLanguage({ languageId: ..., ...language }) instead of language.updateLanguage({ ...language }).

Having it on the entity makes the developer assume that after calling the fn, the state of the object changes but it actually does not do that so we should avoid these pitfalls:

```
const country = Countries.findById({ countryId: "BTC" });
country.isActive; // false
country.updateCountry({ isActive: true });
country.isActive; // still false, but developer might assume it should be true now
```

So in this state it's too inconsistent with the already existing abstraction style.

packages/api/resolvers/mutations/updateAssortmentTexts.js
```diff
@@ -9,13 +9,7 @@ export default function updateAssortmentTexts(
...
-  const changedLocalizations = texts.map(({ locale, ...localizations }) =>
-    assortmentObject.upsertLocalizedText(locale, {
-      ...localizations,
-      authorId: userId,
-    })
-  );
-  return changedLocalizations;
+  const assortment = Assortments.findOne({ _id: assortmentId });
+  if (!assortment) throw new AssortmentNotFoundError({ assortmentId });
+  return assortment.updateTexts({ texts, userId });
```
pozylon, line 14: Contrary to my review comment: This new helper is just fine because if you call assortment.localizedTexts() function after this call it will return the new state and the existing code already uses a collection helper function

packages/api/resolvers/mutations/updateCartItem.js
```diff
@@ -26,27 +26,8 @@ export default function updateCartItem(
...
-  if (configuration !== null) {
-    OrderPositions.updatePosition(
-      {
-        orderId: item.orderId,
-        positionId: itemId,
-      },
-      { configuration }
-    );
-  }
+  if (quantity && quantity < 1)
```
pozylon, line 29: This will lead to quantity 0 not throwing the exception, which is not what you want

packages/api/resolvers/mutations/updateCartItem.js
```diff
@@ -26,27 +26,8 @@ export default function updateCartItem(
...
+    throw new OrderQuantityTooLowError({ quantity });
 
-  return OrderPositions.findOne({ _id: itemId });
+  return order.updateItem({ itemId, quantity, configuration });
```
pozylon, line 32: I think that was already abstracted good enough.

Mikearaya:
I understand your perspective but the main reason I added the function on the object level and not on the prototype level is to avoid making the same `findOne` request to the DB. but i guess the confusion might be worse than the small loss of performance in this case

pozylon:
Hm, where did you save a findOne call? I also checked that and couldn't find an improvement regarding that but if it saves a call to the db we have to look at it more closely

Mikearaya:
I mean the functions inside the package when using prototype function....they have to check if it's a valid ID or not by making a trip to the DB unless they assume that a valid ID will always be provided for them. in that case we can leave findOne call trip to DB. and also there are some edge cases where we might need a trip to DB to validate ID.

pozylon:
> I mean the functions inside the package when using prototype function....they have to check if it's a valid ID or not by making a trip to the DB unless they assume that a valid ID will always be provided for them. in that case we can leave findOne call trip to DB. and also there are some edge cases where we might need a trip to DB to validate ID. like `mutation.registerPaymentCredentials`, `mutation.createFilterOption`

In cases where the current code needs the actual object's helpers it's fine to have another helper I guess

pozylon:
But for most of the mutations with your changes there is the same amount of db trips and not less.

For example makeBase in assortments helpers (https://github.com/unchainedshop/unchained/pull/281/files#diff-f4f2bb0c5ec1ae14bb77be453a6d2bb2e91146dff9b94d62014604c1561cfe4eR546)

You could actually define it as Assortments.makeBase = () ... instead of in the helpers and the amount of db trips doesn't change.

pozylon:
I'd even argue that you will end up with more db trips than required in the future when you have applied changes because of  #246:

Take this change here:
https://github.com/unchainedshop/unchained/pull/281/files#diff-0f44eb781caf5ad56350e5a8dece4862442dbc837f38276288639db754815fbf

Assuming you don't throw if the assortment is not found with the helper way you will still need to fetch the db record in order to call the helper function. In contrast, with a static function you will have less db calls after

packages/api/resolvers/mutations/updateCartItem.js
Mikearaya: quantity 0 will short circuit on the first test since 0 is false.

Mikearaya:
I tried to rearrange method calls that fetch data from DB very gently, I basically changed methods that fewer codes in the codebase relies on and almost only static/prototype methods and not object/helper methods because I think we need to have a discussion on how we should approach them and agree on the general pattern tofollow.
Also to make the review a little less I created another branch that branch out from this to add the resolver optimized checks when checking for the existence of a document i.e use a dedicated method that uses `db.collection.count()` instead of `findOne`. This way we can merge that after we are done with this.

packages/api/resolvers/queries/productCatalogPrices.js
```diff
@@ -1,31 +1,12 @@
...
-          price.countryCode,
-          price.currencyCode,
-          price.maxQuantity,
-          price.amount,
-        ].join('')
-      )
-      .digest('hex'),
-    ...price,
-  }));
+  return product.catalogPrices();
```
pozylon, line 11: 👍

packages/core-assortments/db/helpers.js
```diff
@@ -86,6 +136,14 @@ Collections.Assortments.invalidateFilterCaches = () => {
   });
 };
 
+Collections.AssortmentLinks.findLink = ({ assortmentLinkId }) => {
+  return Collections.AssortmentLinks.findOne(assortmentLinkId);
```
pozylon: Always explicitly use the _id for security reasons, else if somehow assortmentLinkId is set to an object, this fn could leak data:

Change lines like findOne({ _id: assortmentLinkId }) to findOne(assortmentLinkId)

pozylon (changes requested):
packages/api/resolvers/types/**/*.js and packages/api/roles/**/*.js still use findOne and find instead of the dedicated prototype functions

Issue #274, opened 2020-11-30
[FEATURE REQUEST] Extended Work Queue


pozylon:
Currently you cannot filter the Work Queue in the Controlpanel. Sometimes it's important to filter certain work types and status in the Work Queue to dig into problems.

pozylon:
https://git.panter.ch/manul/vps/veloplus-shop/-/issues/836

harryadel:
Cannot access the link. Can you at least pull me a screenshot?

pozylon:
Sorry, the link is not relevant it just says that they want that feature and it makes sense. Basically the work queue should be filterable by type and status.

PR #282, opened 2020-12-22
enable work queue filter #274

Mikearaya:
in adding this feature I had to create a new endpoint that returns the work types registered in the system at the moment, as I couldn't find a better way to achieve this. if you have any suggestions let me know. the other thing is why not filter `status` as I did with `type` instead of using the reduce function which makes it a little complex at first glance https://github.com/unchainedshop/unchained/blob/master/packages/core-worker/director.js#L165.

also `new` status returns both `Allocated` & `New` works and i don't know if that is intended or not.


Mikearaya:
@pozylon since you now can filter the work queue i think only one data table is enough instead of separating them into two 2 (new, recent)

packages/core-worker/director.js
```diff
@@ -168,14 +173,36 @@ class WorkerDirector {
...
+        { $skip: skip },
+        {
+          $project: {
+            _id: '$_id',
+            type: '$_id',
+          },
+        },
+      ])
+      .toArray();
+    return typeList;
```
pozylon, line 202: 🤟

pozylon:
Great! I just have one minor thing to ask:

Is there a special reason you have used ``selectTypes`` instead of just ``types`` for the query parameter name?

pozylon (approved):

pozylon (changes requested):
Oops I was too fast accepting, so the UI still has some issues like lagging when I select/deselect a checkbox and there are multiple warnings thrown when I open browser console. Make sure these are gone. For example sometimes it's not possible to select the types: I get a unique key warning in react:

```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `WorkList`. See https://reactjs.org/link/warning-keys for more information.
```

Then another request:

The schema is not good yet, please:
1. Remove RegisteredWorkTypes type, use already existing "WorkType" enum which is dynamically extended with all possible types. Also it does not make sense to return a type which has an id and a type with the same value
2. selectTypes param is taking [String] which would mean that these params are accepted: null, undefined, [null, undefined], ["SOMETHING"]. Please use [WorkType!] so it's either null or a list of existing types.

Mikearaya:
:) i'll fix that it must have sliped.
the reason i used `selectedTypes` instead of `types` is because  ` static buildQueueSelector({ status = [], workId, ...rest }) ` is used by other part of the codebase that pass it the actually type. plus it's more descriptive :)

pozylon:
Force pushed this, make sure you hard reset when you continue working on this

Mikearaya:
@pozylon i'm almost done with this PR too except bug in infinite-scroll, it keeps on repeating the same row when updating the table (checking and unchecking) status boxes. glance over it if you have the energy please :pleading_face:

pozylon:
Okay so I think it's not quite done and if you change some UI elements maybe that issue with the infinite scroll disappears:

1. You have decided to re-use the SearchDropdown for the types. I think that the SearchDropdown is only really required if you have a fulltext search that needs to emit API calls. I'd go for native semantic UI component because you don't need to refetch workTypes and it's more like a selection instead of a dropdown. So consider refactoring this that workTypes is fetched just once when the UI pops up. See https://codesandbox.io/s/etd1x?module=/example.js
2. Maybe from a UI perspective it makes more sense to see the selectTypes as filters that you can set, so you could also use the default Dropdown with the same multi selection feature like in https://codesandbox.io/s/etd1x?module=/example.js. If you select one or more than it is sent along the query as selectTypes and if no type is selected it will show all types.

pozylon:
Then please have a look at my comment regarding the RegisteredWorkType type, which i think is obsolete:

> Then another request:

You could improve the type stuff like this:

Remove RegisteredWorkTypes type, use already existing "WorkType" enum which is dynamically extended with all possible types. Also it does not make sense to return a type which has an id and a type with the same value
selectTypes param is taking [String] which would mean that these params are accepted: null, undefined, [null, undefined], ["SOMETHING"]. Please use [WorkType!] so it's either null or a list of existing types.

Mikearaya:
Thanks, i've made all the necessary adjustments now, and also I have decided to name the resolver `activeWorkTypes` instead of `workTypes` because it's more descriptive.

packages/api/schema/query.js
```diff
@@ -303,6 +304,10 @@ export default [
       Get a specific work unit by ID
       """
       work(workId: ID!): Work
+      """
+      Get List of currently registered worker plugins
+      """
+      activeWorkTypes: [WorkType]!
```
pozylon, line 310: [WorkType!]!

Mikearaya: I made it like that in the case where there is no worktype so user should expect empty array or array of worktype.

Mikearaya: I know it unlikely there will not be any work but its the right logic.

pozylon (changes requested):
When I add a new job the list does not update correctly, maybe something about the keys?

```
mutation {
  addWork(type: HEARTBEAT) {
    _id
  }
}
```

packages/api/schema/query.js
pozylon: [WorkType]! = [null], I don't think you want to return an array of null

Mikearaya:
@pozylon if you are referring to the work type filter dropdown, did you refresh the page? because that list is filled on the initial render. other than that it's working as expected. can you check again?

Mikearaya:
added 10 seconds of poll interval for work type dropdown so no need of a refresh.

pozylon:
Nood good yet, still same issue, it's not that it's not polling, it was polling before but there are multiple issues in the console, react errors and such and I think there is _id conflicts and load more is broken.

pozylon:
If you add a heartbeat work for example, a new line appears but the lines are rendered wrong (new timestamp not showing).

Mikearaya:
@pozylon can you check out the current state of this PR and let me know if there is anything it's missing. The problem existed before it's just that we haven't noticed it because there were separate tables for each.

I have made work queue query to never stop polling because it needs to update the moment new work is add and that not how infinite scroll works. It requires for load more to be called for it to update.

pozylon:
Maybe the infinite table just does not work well together with polling and the best thing is getting rid of the infinite table

pozylon (approved):
I think we can go with this for the moment, let's see how it behaves in prod and if we have probs we just remove the infinite loading and replace it with some kind of paging

Mikearaya:
Sounds good to me.

PR #284, opened 2020-12-30
Increase token time

harryadel:
Pumped the token validity time to ~30 days.


pozylon:
Good solution using getDate setDate!

PR #289, opened 2021-01-05
New Code Structure

pozylon:
This is a draft of the unchained 1.0 concept, this PR contains:

1. isBase for Language and Country is now configured through ENV LANG and COUNTRY, does not make sense to have in DB and made things complicated and weirdo
2. Added typescript support for various packages
3. Utility functions that were part of unchained:core have been moved to unchained:utils
4. unchained:core is now an umbrella for all core packages that are dependent on each other anyways
5. Core Bookmarks has been refactored to a new context-based code structure without collection helpers
6. A PoC for having Dataloaders in place has been added


Mikearaya:
so far it's really exciting and I can already see the architectural upgrade and clarity for the engine. Analisis shall continue and I think we should also consider documentation and naming conventions we should follow moving forward with this implementation as it's a renovation we will definitely touch all core methods of the engine.
I think anyone that works on any part of the renovation should document all the key services and methods of modules they worked on preferably a [documentation page](https://docs.unchained.shop) worthy format or just a draft on the PR they worked on or something.

pozylon:
Next up we have to POC typescript typing with those modules and services.

@macrozone please look at this PR. We want to move away from helpers, use more pure functions and add architectural support for dataloaders. This touches your latest comment at the Veloplus project where you mentioned that it should be more easy for developers to build API's. With this new way, all unchained core functionality could be accessible through a public interface sitting in the apollo context. That way you can basically add a resolver and then use VS auto completion to pull together the data that is needed for your case instead of importing packages and use internal undocumented functions like now.

pozylon:
We also try to refactor the code so it's more of a split between read and write models following some research around CQRS

packages/core-bookmarks/bookmarks.js
```diff
@@ -1,7 +1,50 @@
...
 export * from './db/collections';
-export default () => {
+
+export const services = {};
+
+// eslint-disable-next-line
+export default (config) => {
   runMigrations();
+  // eslint-disable-next-line
+  return (req) => {
```
macrozone, line 14: req is unused.

I also do not understand, why it is wrapped in a req closure. I guess the idea is that the module is re-initialized for every request?

packages/core-bookmarks/bookmarks.js
```diff
@@ -1,7 +1,50 @@
...
+    const bookmarksByQueryLoader = new DataLoader(async (queries) => {
+      const results = Bookmarks.find({
+        $or: queries,
+      }).fetch();
+      return queries.map(
+        (key) =>
+          results.find(
+            (result) =>
+              result.userId === key.userId && result.productId === key.productId
+          ) || null
```
macrozone, line 24: `?? ` is probably better here, because with || you will convert every falsy value to null

packages/core-bookmarks/bookmarks.js
```diff
@@ -1,7 +1,50 @@
...
+      );
+    });
+    const bookmarkByIdLoader = new DataLoader(async (keys) => {
```
macrozone, line 27: ids would be a better name than keys

packages/core-bookmarks/bookmarks.js
```diff
@@ -1,7 +1,50 @@
...
+      const results = Bookmarks.find({
+        _id: {
+          $in: keys,
+        },
+      }).fetch();
+      return keys.map(
+        (key) => results.find((result) => result._id === key) || null
```
macrozone, line 34: careful here with the || 

altough of course there won't be any falsy values as its either an object or undefined.

packages/core-bookmarks/bookmarks.js
```diff
@@ -1,7 +1,50 @@
...
+      findBookmarkByUserIdAndProductId: async ({ userId, productId }) =>
+        bookmarksByQueryLoader.load({ userId, productId }),
+      findBookmarkById: async (bookmarkId) =>
+        bookmarkByIdLoader.load(bookmarkId),
+      removeBookmarkById: async (bookmarkId) => {
+        return Bookmarks.removeBookmark({ bookmarkId });
+      },
+      createBookmark: async (bookmarkData) => {
+        return Bookmarks.createBookmark(bookmarkData);
+      },
```
macrozone, line 47: you could remove `bookmark` from the function names as they exist on the `bookmark` module

packages/core-bookmarks/bookmarks.js
macrozone: well i guess that is the idea for dataloader: you cache per request. so nevermind this comment

packages/api/api.js
```diff
@@ -25,28 +25,41 @@ const defaultContext = (req) => {
   return { remoteAddress };
 };
 
-export const createContextResolver = (context) => async ({ req }) => {
-  const userContext = await getUserContext(req);
+export const createContextResolver = (context, unchained) => async ({
+  req,
+}) => {
+  const unchainedContext = {
+    modules: await unchained?.instantiateLoaders(req),
```
macrozone, line 32: `instantiateModules` would be better name.

macrozone:
i like this direction. 

One thing to consider is, where the unchained business logic actually is. i think its good to have a `unchained` object around that does all the business logic and that this is also available in resolvers, so that resolver have no business logic at all. Not sure if data-loaders should be part of that though, i think they are an optimisation for resolvers? So they would already use this "unchained" api object as data-source. They are currently named "modules", which is maybe too ambiguous. 

What maybe helps is to sketch out the different "layers" (resolvers <-> dataloaders <->  acl <->  businesslogic <->  database-access ?) and how you modify these layers (As unchained core devs and consumers of unchained).

I also would start to add typescript now, better sooner than later, because it will become very complicated with these module injections and you will be much faster with refactoring if you know what actually exist on the resolver context. 
you can easily opt-in per file. I would start to type the graphql "Context", because it will probably contain most data-types directly or indirectly.  

I guess this type will contain a lot of generics in the future so that its easy for devs to extend the core api!

pozylon:
@schmidsi Can you have a look at this PR too, It's crucial that we have everybody onboard with these fundamental decisions, be it open source devs, customers, partners or employees of Unchained Commerce.

pozylon:
> Not sure if data-loaders should be part of that though, i think they are an optimisation for resolvers?

You're propably right about this, wrapping business logic or db functions in data-loaders is something that is actually only important for GraphQL at the moment so they'd have to live separate from the modules.

Layers from top to bottom:
1. Platform
2. API
3. Business Logic (Unchained Context)
4. DB

The responsibility of the Platform layer is:
1. Convenience Function "startPlatform" to boot up a default Unchained environment
1. Load and configure core modules and custom modules
2. Load and configure auth and user roles
3. Load and configure messaging templates
4. Load and configure the API

The responsibility of the API layer is:
1. Permissions
2. Marry the business logic layer with the GraphQL schema (connect read & write, map errors)
3. Allow extending the default GraphQL server: Add more data/functions to the context, Add/Extend GraphQL Schema types, Add/Overwrite Resolvers, Extend ACL Permissions

- Business logic should separate read and write models where it is easily possible and makes sense and exposes a global Unchained Context with all modules and services. modules and services differentiate each other by modules beeing core- prefixed packages that provide DB access and services beeing "cross-package" business logic like pricing, cart, discounting.

- DB Layer for the moment is basically tied to MongoDB, no ORM planned yet. we plan to switch from Meteor Mongo to mongodb-native though.

pozylon:
Inspired by: https://strapi.io/documentation/developer-docs/latest/concepts/services.html#concept

harryadel:
TBH, I don't have much experience in GraphQL migrations especially one of this caliber so I can't weight in much. My only concern is that wouldn't lumping many services under `modules` namespace would complicate things down the road?

pozylon:
I've added typescript and took all of your comments into account so we can have another iteration over this draft

packages/core-bookmarks/bookmarks.ts
```diff
@@ -21,14 +21,24 @@ export const services = {
 };
 
 export interface UnchainedBookmarkAPI {
-  findByUserId({ userId: String }): Promise<Array>;
+  findByUserId(userId: string): Promise<Array>;
```
macrozone, line 24: That should Probably be `Promise<User>`

packages/core-bookmarks/bookmarks.ts
```diff
@@ -21,14 +21,24 @@ export const services = {
...
+  findByUserIdAndProductId({ userId: string, productId: string }): Promise<any>;
```
macrozone, line 25: don't use `any`. Shoud probably be `Promise<Bookmark[]>` or `Promise<Bookmark>` as there can only be one

packages/core-bookmarks/bookmarks.ts
```diff
@@ -21,14 +21,24 @@ export const services = {
...
+  findById(bookmarkId: string): Promise<any>;
```
macrozone, line 26: same same

packages/api/resolvers/mutations/createBookmark.js
```diff
@@ -1,23 +1,35 @@
...
-  return Bookmarks.createBookmark({ userId: foreignUserId, productId });
+  const bookmark = await modules.bookmarks.findByUserIdAndProductId({
+    productId,
+    userId: foreignUserId,
+  });
+
+  if (bookmark)
+    throw new BookmarkAlreadyExistsError({ bookmarkId: bookmark._id });
+
+  const bookmarkId = await modules.bookmarks.create({
```
schmidsi, line 30: I like this API

packages/api/resolvers/types/bookmark.ts
```diff
@@ -0,0 +1,11 @@
+import { Users } from 'meteor/unchained:core-users';
+import { Products } from 'meteor/unchained:core-products';
+
+export default {
+  product(obj) {
+    return Products.findOne({ _id: obj.productId });
```
schmidsi, line 6: Shouldn't you also use `modules.bookmarks...` here?

schmidsi (approved):

schmidsi:
Tests are also green on my computer. I think we need to merge here fast so that other changes like in #253 can benefit from the new setup/structure.

It's probably not perfect yet (with `any` types), but it is good to go in my opinion. I don't see any big issues that we regret later. Although, I did not look super deep into it. I can give a better feedback as soon as I tried to write/rewrite a module by myself.

schmidsi:
There are type errors in the code now. I added typechecks to the test pipeline so CI fails now.

schmidsi:
I fixed the bookmarks.ts typecheck errors. But there are still some import errors.

Just as a motivation for types:
Screenshot 2021-01-09 at 21 08 18

During fixing the type errors, I discovered that `Bookmarks.update` returns a number and not an object, like the original interface suggested. Having the Typescript compiler pointing to this discrepancy now can save a lot of headaches later.

packages/api/resolvers/types/bookmark.ts
pozylon: To find products?

packages/api/api.ts
```diff
@@ -0,0 +1,76 @@
...
+}): Promise<UnchainedServerContext> => {
+  const loaders = await instantiateLoaders(req, unchained);
+  // const intermediateContext: Partial<UnchainedServerContext> = {
+  //   ...unchained,
+  //   ...loaders,
+  // };
+  const userContext = await getUserContext(req /* intermediateContext */);
+  const localeContext = await getLocaleContext(req);
+  return {
+    ...unchained,
```
schmidsi, line 48: Hmm, I'm not sure about this. It's a bit a lot of magic-merges. This can lead to problems if for example `loaders` also has a key `modules`. Why not making just:

```ts
return {
  unchained,
  loaders,
  user: userContext,
  locale: localeContext
  // version is part of `unchained`
}

pozylon: Also thought about it but then If I change this part I'll break all explicit and implicit (helper functions) resolvers so I didn't do it for now.

pozylon: This will be so much easier to refactor once we have all resolvers and fields explicit and typed

schmidsi: Ok, that makes a lot of sense.

PR #292, opened 2021-01-07
Typescript

schmidsi:
This are my first steps into making the Engine Typescript ready. Linting goes through and typechecking in VS code worked for my testfile `core-worker/plugins/heartbeat.ts`.  But somehow, the tests do not run at all on my computer.

It is only a draft because the tests didn't went through yet.


.eslintrc
```diff
@@ -28,6 +34,22 @@
...
-    ]
-  }
+    ],
+    "no-unused-vars": "off",
+    "@typescript-eslint/no-unused-vars": "warn",
+    "@typescript-eslint/no-empty-function": "warn",
+    "@typescript-eslint/explicit-module-boundary-types": "off",
+    "@typescript-eslint/no-var-requires": "off"
+  },
+  "overrides": [
```
schmidsi, line 44: This is a cool pattern that I discovered: We can introduce more strict linting rules for .ts files. So the code quality increases by migrating to Typescript not only through the types but also through the linting. But the old code just works as before.

packages/api/api.js
```diff
@@ -4,10 +4,10 @@ import createGraphQLServer from './createGraphQLServer';
 import createBulkImportServer from './createBulkImportServer';
 import { configureRoles } from './roles';
 
-export hashPassword from './hashPassword';
-export getCart from './getCart';
-export evaluateContext from './evaluateContext';
-export filterContext from './filterContext';
+export { default as hashPassword } from './hashPassword';
```
schmidsi, line 7: This is needed for the Typescript parser

packages/core-worker/plugins/heartbeat.ts
```diff
@@ -1,15 +1,25 @@
...
 
-class Heartbeat extends WorkerPlugin {
+type Arg = {
+  wait?: number;
+  fails?: boolean;
+};
+
+type Result = Arg;
+
+class Heartbeat extends WorkerPlugin<Arg, Result> {
```
schmidsi, line 22: Example of a correctly typed worker-plugin

packages/core-worker/types.d.ts
```diff
@@ -0,0 +1,16 @@
+declare module 'meteor/unchained:core-worker' {
```
schmidsi, line 1: How to type `import * from 'meteor/unchained:core-worker';` imports

schmidsi:
@pozylon Maybe you can merge these changes to https://github.com/unchainedshop/unchained/tree/ts-ausflug

.eslintrc
pozylon: Nice

pozylon:
I'll merge it into the bookmark PoC then fix tests

harryadel:
Meteor Roles PR contains some typescript setup, maybe we can use it as a starting point to avoid replicating work.

https://github.com/unchainedshop/unchained/pull/253

pozylon:
Part of bookmarks poc now so we can discuss future direction there

PR #299, opened 2021-01-13
Remove not found error

Mikearaya:
i've left `PaymentProviderNotFoundError` from `query.signPaymentProviderForCredentialRegistration` because it's almost a mutation and returning null might not be enough. plus anyone who calls this endpoint should be expected to provide a correct `paymentProviderId`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
   },
   "scripts": {
     "lint": "eslint packages tests --cache --fix --ext=js --ext=ts",
-    "pretest": "npm run lint --silent && npm run typecheck",
     "lint:ci": "eslint packages tests --ext=jsx --ext=js",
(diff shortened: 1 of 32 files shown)


pozylon:
Great! Maybe that's a sign we have to change it to Mutation.signPaymentProvider... what do you think ? It sure invokes side-effects

Mikearaya:
Yes but how big is the impact of this breaking change and should we introduce it now?

pozylon (approved):
I leave that up to you, code looks good for merge

Mikearaya:
so far weng and publicare are not using it so I've decided to change it to a mutation. it would be great if you can confirm veloplus and others.
I'm merging this

Issue #300, opened 2021-01-19
signPaymentProviderForCredentialRegistration should be a mutation


Mikearaya:
currently, `signPaymentProviderForCredentialRegistration` is exposed as a query but it should be a mutation instead because it modifies the application state

PR #301, opened 2021-01-19
Change signPaymentProviderForCredentialRegistration 

Mikearaya:
…query and update tests #300


pozylon:
Two things:

1. I think the actual signing for payment and not for credential registration should also be moved to a mutation, this is OrderPaymentGeneric.sign to Mutation.signPaymentProviderForCheckout
2. After looking at the changes I think we could provide the old queries/fields still with a deprecation through @deprecated, that would not immediately break the storefronts that build on this.

package.json
```diff
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "lint": "eslint packages tests --cache --fix --ext=js --ext=ts",
-    "pretest": "npm run lint --silent && npm run typecheck",
```
pozylon, line 23: Sure?

packages/api/locale-context.ts
```diff
@@ -7,10 +7,12 @@ import {
 
 import { log } from 'meteor/unchained:core-logger';
 import { Countries } from 'meteor/unchained:core-countries';
-import { Locales } from 'locale';
 import { Languages } from 'meteor/unchained:core-languages';
 import LRU from 'lru-cache';
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const locale = require('locale');
```
pozylon, line 14: Use import

packages/api/schema/query.js
```diff
@@ -308,6 +301,16 @@ export default [
...
       """
       activeWorkTypes: [WorkType!]!
+
+      """
+      Sign a generic payment provider for registration
+      """
+      signPaymentProviderForCredentialRegistration(
+        paymentProviderId: ID!
+      ): String
+        @deprecated(
```
pozylon, line 311: OrderPayment.sign should be marked deprecated too

packages/core-orders/db/order-payments/helpers.js
```diff
@@ -32,6 +32,12 @@ OrderPayments.helpers({
   normalizedStatus() {
     return objectInvert(OrderPaymentStatus)[this.status || null];
   },
+  sign({ transactionContext }) {
```
pozylon, line 35: DRY: Reuse in order-payment-generic.js GraphQL type resolver

packages/api/resolvers/mutations/signPaymentProviderForCheckout.js
```diff
@@ -0,0 +1,26 @@
...
+  const orderPayment = OrderPayments.findPayment({ orderPaymentId });
+  if (!orderPayment) throw new OrderPaymentNotFoundError({ orderPaymentId });
+  const providerType = orderPayment?.provider()?.type;
+  if (providerType !== PaymentProviderType.GENERIC)
+    throw new OrderPaymentTypeError({
+      orderPaymentId,
+      received: providerType,
+      required: PaymentProviderType.GENERIC,
+    });
+  return orderPayment.sign({ transactionContext });
```
pozylon, line 25: The current implementation of OrderPayment.sign throws a OrderPaymentConfigurationError if signature could not be created.

packages/api/schema/mutation.js
```diff
@@ -880,6 +880,20 @@ export default [
...
+      Sign a generic payment provider for registration
+      """
+      signPaymentProviderForCredentialRegistration(
+        paymentProviderId: ID!
+      ): String
+
+      """
+      Sign a generic order payment
+      """
+      signPaymentProviderForCheckout(
```
pozylon, line 893: Maybe we should use a non-optional String as return type. If a payment provider does not support signing we could throw that as error like "This Payment Provider does not support signing transactions" or something like that.

pozylon (changes requested):

pozylon:
Cool, I think we are almost there.

changelog.md
```diff
@@ -3,13 +3,16 @@
...
+
 - [api] `Mutation.setBaseLanguage` removed, base language now set through env `LANG`
 - [api] `Mutation.setBaseCountry` removed, base language now set through env `COUNTRY`
 - [api] Remove OTP based access to an order
 
 ## Major
+
 - [api] Add Assortment.childrenCount to get a number of child assortments
 - [api] New Query.activeWorkTypes to query for all active work types without introspection
+- [api] New mutation.signPaymentProviderForCheckout to sign generic order payment directly
```
pozylon, line 15: Move to a new top level version vNEXT

pozylon (commented):
Forgot the changelog

package.json
Mikearaya: `dataloader` and `locale` fail with pretest and I removed that for other test but forgot

packages/api/locale-context.ts
Mikearaya: They fail on pretest when using import

changelog.md
Mikearaya: ok, was intending this to be included with v0.61 :)

packages/api/schema/mutation.js
Mikearaya: I think we can safely make it return a non-optional String because it's already making that check inside the resolver and `OrderPayment.sign` is called in GENERIC type only so...

PR #309, opened 2021-01-30
Hot fix user seed error

Mikearaya:
I rebased on 0.61.x so everything I'm working on doesn't get funny when merging. but I encountered this error when I tried to delete the `meteor` and create a new one. this was the error

Screenshot from 2021-01-30 21-12-09

it has to do with the `createUser` function sending `emails` array instead of `email` but set enrollment is actually doing this
```
await accountsPassword.sendEnrollmentEmail(userData.email);
```


pozylon:
Good catch, but I'd rather only leave one option to set emails when using createUser, going to apply that to your PR and merge. Using "emails" circumvents inner logic of accountsjs and for example the lowercasing.

Issue #214, opened 2020-09-09
Get rid of "littledata:synced-cron"


pozylon:
- [ ] Remove all traces of littledata:synced-cron
- [ ] Remove the CronWorker and improve the IntervalWorker so it's a drop-in replacement

pozylon:
The cool thing about synced-cron is that it's synced :D

PR #310, opened 2021-02-01
getting rid of littledata:synced-cron #214

pozylon:


packages/platform/setup-workqueue.js
```diff
@@ -20,14 +19,9 @@ export default ({ cronText }) => {
...
-  //   new IntervalWorker({
-  //     WorkerDirector,
-  //   })
-  // );
   if (cronText) {
     handlers.push(
-      new CronWorker({
+      new IntervalWorker({
         WorkerDirector,
         cronText,
```
Mikearaya, line 26: Has the format change, is it Linux corn text now.

Mikearaya (approved):
I see you activated the interval work. looks good to me

packages/platform/setup-workqueue.js
pozylon: No I think it's the same like before, isn't it?

pozylon: You're right it's actually not cron it's kind of a text parser (http://bunkat.github.io/later/parsers.html#text) so the name is missleading, we should support both

pozylon: I saw we have this problem at different places there is even a setting "autoSchedulingCronText" that is actually not cron it's later.js text expression.

pozylon:
@Mikearaya I took a radical approach and removed cronText completely, the name was so missleading and deprecating it would not have helped there.

pozylon:
At least now you could use linux cron tab expressions for the interval worker (https://bunkat.github.io/later/parsers.html#cron)

Issue #294, opened 2021-01-12
Event Tracking Module that allows to connect external Analytics services


schmidsi:
The idea here is, that every Unchained-module has a set of events that are emitted.

Events should be typed. This could be one possibility to type events: https://basarat.gitbook.io/typescript/main-1/typed-event

The worker module has already some sort of events, but it is not perfect:
https://github.com/unchainedshop/unchained/blob/master/packages/core-worker/director.js#L12-L21
https://github.com/unchainedshop/unchained/blob/master/packages/core-worker/director.js#L125

# Step 1: Define events
So for every module, we define a set of events and type them properly:

For example: `core-product` has the following types according to https://docs.unchained.shop/api/entities/#endpoint:
- CREATE
- UPDATE
- DELETE

Or `core-orders` has:
- ADD_PRODUCT
- ADD_DISCOUNT
- CHECKOUT
- CONFIRMED
- ...
(these might not be the best names, I'm open for suggestions here)

# Step 2: Emit events
Then the module's director or module's main collection is extended to emit these events when they occur.

# Step 3: Browsing-events module
To actually be able to track also standard events like "PAGE_VIEW", we need a module that defines these events and exposes GraphQL mutations to actually trigger these events from the front-end.

# Step 4: Example connector to Matomo
Now that we have all modules emitting events, we can now write a conntector-module which listens for these events and puts them in an external service. A good starting point would be Matomo.

# Workflow

* Do Step 1 and 2 ony for `core-orders` and open a Pull request so we all can discuss the introduced patterns
* Do Step 3
* Do Step 4 -> We can ship this feature the for Publicare
* Do Step 1 and 2 for the other modules

Mikearaya:
Making the modules main collection define event is not a good pattern in my opinion and in those cases defining event emitter class on its on might be good. Mainly it will be decoupled from the data.
On the other hand modules that have director class already can define events such as the work module.
So we can either follow the same pattern of defining event emitter per module or define emitter for modules that don't have directors and add event emitter inside each modules director if they have one.

pozylon:
I agree with 1, 3, 4 and the Workflow but like @Mikearaya I think we have to talk about 2 in more details.

I'd go even further and try to encapsulate the event emitting into it's own package and make it a single function call like when you log something. So rather than re-implementing a very basic Event Emitting pattern I'd add an abstraction layer.

[...]

PR #316, opened 2021-02-07
order events tracking #294

Mikearaya:
First step: Typing Order event emitter and  Declaring all the possible events that could be emitted from the module


schmidsi (commented):
@Mikearaya Why is there a new branch: https://github.com/unchainedshop/unchained/compare/core-events ?

PR #324, opened 2021-02-16
Core events

Mikearaya:


Mikearaya:
@schmidsi this is the branch that one had a conflict plus it's not the module approach so i created a new branch. that one is deleted now tho

packages/core-events/director.ts
```diff
@@ -0,0 +1,41 @@
...
+  static emit(eventName: string, payload: any): void {
+    EventDirector.adapter.publish(eventName, payload);
+  }
+
+  static subscribe(eventName: string, callBack: () => void): void {
+    EventDirector.adapter.subscribe(eventName, callBack);
+  }
+}
+
+class EventPlugins extends EventAdapter {
```
schmidsi, line 25: So this is the actual adapter using Node's EventEmitter?

packages/core-events/director.ts
```diff
@@ -0,0 +1,41 @@
+import EventEmitter from 'events';
+
+export abstract class EventAdapter {
+  public abstract publish(eventName: string, payload: any): void;
```
schmidsi, line 4: I'd mimik node's terminology:

- emit
- on
- once

package.json
```diff
@@ -20,7 +20,6 @@
   },
   "scripts": {
     "lint": "eslint packages tests --cache --fix --ext=js --ext=ts",
-    "pretest": "npm run lint --silent && npm run typecheck",
```
schmidsi, line 23: Why?

packages/core-events/director.ts
```diff
@@ -0,0 +1,41 @@
...
+    EventDirector.adapter = adapter;
+  }
+
+  static emit(eventName: string, payload: any): void {
+    EventDirector.adapter.publish(eventName, payload);
+  }
+
+  static subscribe(eventName: string, callBack: () => void): void {
+    EventDirector.adapter.subscribe(eventName, callBack);
+  }
```
schmidsi, line 22: We also need a "register" event hook.

packages/core-events/director.ts
Mikearaya: yes, default support

Mikearaya:
@schmidsi @pozylon 
so far implementing order event emitting I've identified this event and its corresponding payloads

**event**                                   **payload**
```
ORDER_CHECKOUT =>  current order object
ORDER_ADD_PRODUCT => Order position
ORDER_ADD_DISCOUNT => reservedDiscount
ORDER_CONFIRMED => current order object
ORDER_FULLFILLED => current order object
ORDER_UPDATE_PAYMENT => current order payment object
ORDER_UPDATE_DELIVERY => current order delivery object
ORDER_SIGN_PAYMENT => { orderPayment,  transactionContext }
ORDER_REMOVE => { orderId }
```

inputs....looks like there needs to be a lot more decision-making required from the group per module on what the emitted format should be and the content of the event.
I've implemented the above events and we can discuss them over this

pozylon:
ORDER_UPDATE in general is missing (changed contact, billing, meta, ...) + ORDER_SET_PAYMENT_PROVIDER + ORDER_SET_DELIVERY_PROVIDER

pozylon:
I wonder what's better when naming events:
ORDER_UPDATED or ORDER_UPDATE?

I think the Event Payload for orders should mostly just contain id's rather than actual objects. The code that hooks into an event must have access to the unchained "context" that is also passed to apollo.

So it's like onEvent({ orderId }, futureSuperContext) => futureSuperContext.modules.orders.findOrder({ orderId })... to get the object

schmidsi:
> I wonder what's better when naming events:
> ORDER_UPDATED or ORDER_UPDATE?

Usually, event names are present tense. See: https://nodejs.org/api/http.html#http_class_http_clientrequest

> I think the Event Payload for orders should mostly just contain id's rather than actual objects. The code that hooks into an event must have access to the unchained "context" that is also passed to apollo.
> 
> So it's like onEvent({ orderId }, futureSuperContext) => futureSuperContext.modules.orders.findOrder({ orderId })... to get the object

Hmm, I had in mind that the event payload should be a JSON (no functions) and should contain all relevant data. Relevant is: So that a read only observer that observes all events of the system should be able to rebuild the state of the system at any point in time (time travel).

If you put only the ID and you query for this id later (from next tick up to minutes, hours or days), the information might already have changed.

pozylon:
Yeah I see what you mean. 👍Objects would allow for retrospectivity which is what we need.

Mikearaya:
based on this event names should be CamelCase but I think we are following our own standards when we are using caps. when it comes to the name I also think they should define the action performed (ORDER_UPDATED) rather than the action requested (ORDER_UPDATE).
like Simon said I think limiting the type of user that can read the actual data by only emitting id and letting the user query it from unchained might not be necessary and we could send the actual object the event-related to. the only benefit I see from it is controlling restricted users otherwise anyone can subscribe to an event and get the information they want.

Mikearaya:
forgot to put the link https://basarat.gitbook.io/typescript/main-1/typed-event

schmidsi:
> based on this event names should be CamelCase but I think we are following our own standards when we are using caps. when it comes to the name I also think they should define the action performed (ORDER_UPDATED) rather than the action requested (ORDER_UPDATE).

I think we should follow community standards if possible. But if you have a really good argument, why we should use UPPER_CASE and past tense, go on.

> like Simon said I think limiting the type of user that can read the actual data by only emitting id and letting the user query it from unchained might not be necessary and we could send the actual object the event-related to. the only benefit I see from it is controlling restricted users otherwise anyone can subscribe to an event and get the information they want.

Access control happens at the resolver level. Below, everything is open.

> ... we could send the actual object the event-related to.

... after the change.

pozylon:
Upper case kind of signals to me that it is an enum of some kind and it gets highlighted by the IDE as such. And it's usually the style when you define types and enums in GraphQL. Event names for me are to be handled like enums.

PascalCase is obviously also an option 😂 no just joking I don't care. I could personally live with present style because order_checkedout sucks compared to order_checkout

pozylon:
@schmidsi which community standards about event names do you refer to?

Mikearaya:
@pozylon  https://basarat.gitbook.io/typescript/main-1/typed-event

schmidsi:
> @schmidsi which community standards about event names do you refer to?

I just quickly looked over the event names here: https://nodejs.org/api/http.html#http_class_http_clientrequest I don't know if there is a guide. Personally, I like the UPPER_CASE also. Present tense is just shorter.

Mikearaya:
@schmidsi @pozylon 1 & 2 done ready for review and move on to 3

Mikearaya:
Failing test are because of rebase with master

packages/api/resolvers/mutations/pageView.js
```diff
@@ -0,0 +1,22 @@
+import { log } from 'meteor/unchained:core-logger';
+import { emit } from 'meteor/unchained:core-events';
+
+export default (
+  root,
+  { path },
```
schmidsi, line 6: Please add referrer here also

schmidsi: https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer

packages/core-events/core-events-tests.js
```diff
@@ -0,0 +1,11 @@
+// Import Tinytest from the tinytest Meteor package.
```
schmidsi, line 1: Delete that file

packages/core-events/director.ts
schmidsi: @Mikearaya ??

schmidsi (approved):

Mikearaya:
@schmidsi ? you suggest interface over abstract class ?

schmidsi:
> @schmidsi ? you suggest interface over abstract class ?

No, I meant following the node.js standard and use `emit` instead of `publish`. And `on` instead of `subscribe` and `off` instead of `unsubscribe` and `once` for only get the next event'

pozylon (commented):
From what I see now, I'd suggest some more tasks. I know that most of the following is not part of the steps in the issue but I think that stuff should be part of such a feature, no?

Admin Features:
- Queries that allow to query events (only admin by default)
- Controlpanel Views to view events (only admin by default)
- (GraphQL subscription to subscribe to the event stream (only admin by default))

Additional plugins:
- Out-of-the-box redis pub/sub plugin because we already use redis heavily for apollo and thumbor caching

And:
- EventDirector should log event emit and subscription with verbose or debug log level
- Matomo should log fetch url's together with the event it subscribed with debug log level
 
And I'm wondering if PAGE_VIEW also catches the browser version from req?

packages/core-orders/orders.js
```diff
@@ -7,9 +8,32 @@ export * from './db/order-documents';
...
+  'ORDER_REMOVE_DISCOUNT',
+  'ORDER_DELIVER',
+  'ORDER_PAY',
+];
 
 export default () => {
   // configure
   Orders.invalidateProviders();
   runMigrations();
+  registerEvent(ORDER_EVENTS);
```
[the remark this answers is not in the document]
pozylon: because it takes an array

pozylon:
Else, structure wise I have nothing more to say, looks very good!

Mikearaya:
Thanks i can work with this

schmidsi:
> And I'm wondering if PAGE_VIEW also catches the browser version from req?

Technically it should be possible to get the userAgentString from req.

Mikearaya:
@pozylon almost all the requirements are met except **GraphQL subscription to subscribe ** the reason being not every event package ave streaming eg. nodejs EventEmitter, while redis does so currently i can see how to implement this with out having an Apollo server dependency in the engine and can't  abstract EventDirecror enough to accommodate every type of event packages 

AsyncIterator is a good candidate.
I would appreciate it if you can give me your thoughts on this.
Thanks

pozylon:
@Mikearaya I think we can scrub the requirement, I can't really see why we should allow to tail events with the standard. If there is any super-realtime case where such things are needed, we can add it in that project.

index.js
```diff
@@ -0,0 +1,7 @@
+const redis = require('redis');
```
pozylon, line 1: Is this file for test purposes? We could write an integration test for the redis plugin by extending codeship's codeship-services.yml:

here is an example from https://docs.cloudbees.com/docs/cloudbees-codeship/latest/pro-builds-and-configuration/services on how to run a sidecart postgres, the same way we could side-load redis.
```
app:
  build:
    image: codeship/app
    dockerfile: Dockerfile
  links:
    - postgres
postgres:
  image: healthcheck/postgres:alpine
```

package.json
```diff
@@ -20,7 +20,7 @@
...
```
pozylon, line 23: Disabled linting?

pozylon (changes requested):
We need a big fat readme entry in vNext + good docs from the beginning on how to use this feature.

pozylon:
This is going to be a big fat feature to build the future on! :)

pozylon:
@Mikearaya Please go through the changes, resolve everything and rebase on master so we can merge this in

package.json
pozylon: @Mikearaya ping

index.js
pozylon: @Mikearaya ping

Mikearaya: @pozylon noted

Mikearaya: it was for test

packages/core-orders/db/orders/helpers.js
```diff
@@ -673,7 +691,9 @@ Orders.createOrder = async ({
     currency,
     countryCode,
   });
-  return Orders.findOne({ _id: orderId }).initProviders();
+  const order = Orders.findOne({ _id: orderId });
+  emit('ORDER_CREATE', { order });
```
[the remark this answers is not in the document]
Mikearaya: still returning that object

Mikearaya:
I've added events to some of the other modules now. about testing Redis, yes I can get it up and running in CI but the flow is different because it's a fire and forget where do you test for the event? but this can wait for the next round.

pozylon:
@Mikearaya I'll do the last minor fixes and merge it, great work!

pozylon (approved):

Issue #322, opened 2021-02-09
Test for product price ranges


Mikearaya:
Add tests for the new price range fields of a product
* `ConfigurableProduct.simulatedPriceRange` 
* `ConfigurableProduct.catalogPriceRange`
* `SimulateProduct.leveledCatalogPrices` 
* `PlanProduct.leveledCatalogPrices`

PR #329, opened 2021-02-25
Price resolver test

Mikearaya:
#322

### DIFF
--- a/packages/core-products/db/products/helpers.js
+++ b/packages/core-products/db/products/helpers.js
@@ -60,12 +60,38 @@ const getPriceRange = (prices) => {
 
   return {
     minPrice: {
+      _id: crypto
+        .createHash('sha256')
+        .update(
+          [
+            this._id,
+            Math.random(),
+            min?.isTaxable,
+            min?.isNetPrice,
+            min?.amount,
+            min?.currencyCode,
+          ].join('')
+        )
+        .digest('hex'),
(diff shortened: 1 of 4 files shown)


Mikearaya:
@pozylon i found some errors related with `_id` not set on the newly added price revolvers and I've fixed it on this PR

pozylon:
When you're done with the tests and Codeship runs through you can merge right away

packages/core-products/db/products/helpers.js
```diff
@@ -60,12 +60,38 @@ const getPriceRange = (prices) => {
 
   return {
     minPrice: {
+      _id: crypto
+        .createHash('sha256')
+        .update(
+          [
+            this._id,
+            Math.random(),
```
pozylon, line 68: If you use Math.random here, there is no need for any hash creation. The hashing is thought to make the _id stay the same when the price does not change

pozylon (changes requested):

packages/core-products/db/products/helpers.js
Mikearaya: @pozylon id for min and max price will be the same if i don't find a differentiator. that's why i used it

pozylon: Then use "min" and "max" please. Math.random should never be part of an _id hash

Issue #312, opened 2021-02-04
`usersCount` query


schmidsi:

schmidsi:
As proposed by @macrozone

macrozone:
its probably good to have a *Count query for every list (productsCount, assortmentsCount, etc)

they should take the same args as their list counterparts, minus sort and pagination

PR #342, opened 2021-03-27
Collection counters

Mikearaya:
for #312

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "lint": "eslint packages tests --cache --fix --ext=js --ext=ts",
-    
+    "pretest": "npm run lint --silent && npm run typecheck",    
     "lint:ci": "eslint packages tests --ext=jsx --ext=js",
     "test": "jest --runInBand --detectOpenHandles --forceExit tests/*",
     "test:ci": "npm run test:integration && npm run test:packages",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "lint": "eslint packages tests --cache --fix --ext=js --ext=ts",
-    "pretest": "npm run lint --silent && npm run typecheck",
+    
     "lint:ci": "eslint packages tests --ext=jsx --ext=js",
     "test": "jest --runInBand --detectOpenHandles --forceExit tests/*",
     "test:ci": "npm run test:integration && npm run test:packages",
--- a/packages/core-assortments/db/helpers.js
+++ b/packages/core-assortments/db/helpers.js
@@ -110,6 +110,32 @@ Collections.Assortments.findAssortments = ({
   return Collections.Assortments.find(selector, options).fetch();
 };
 
+Collections.Assortments.count = async ({
+  tags,
+  slugs,
+  includeInactive,
+  includeLeaves,
+}) => {
+  const selector = {};
+
+  if (slugs?.length > 0) {
+    selector.slugs = { $in: slugs };
+  } else if (tags?.length > 0) {
+    selector.tags = { $all: tags };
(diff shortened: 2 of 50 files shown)


packages/core-assortments/db/helpers.js
```diff
@@ -110,6 +110,32 @@ Collections.Assortments.findAssortments = ({
   return Collections.Assortments.find(selector, options).fetch();
 };
 
+Collections.Assortments.count = async ({
```
pozylon, line 113: Super nice you have added these count functions to the collections but please extract the selector building code to minimize duplication and reduce likelihood of wrong counts when just one selector got changed in one function.

Kind of like:
```
const buildFindSelector = (props) => ({ isActive: ..., ... }); // contains selector building code of findAssortments
Assortments.findAssortments = (props) { sort=..., Assortments.find(buildFindSelector(props), sort).fetch() }
Assortments.count = (props) { Assortments.find(buildFindSelector(props)).count() }
```

pozylon (changes requested):
I found some already existing count fields and I think we should unify the naming and deprecate the old ones for these:

```
ype ProductCollection {
      totalCount: Int!
      filteredCount: Int!
      items: [Product!]!
    }

type LoadedFilterOption {
      filteredProducts: Int!
      definition: FilterOption!
      isSelected: Boolean
    }

type LoadedFilter {
      examinedProducts: Int!
      filteredProducts: Int!
      definition: Filter!
      isSelected: Boolean
      options: [LoadedFilterOption!]
    }

"""
    Search result
    """
    type SearchResult {
      totalProducts: Int!
      filteredProducts: Int!
      filters: [LoadedFilter!]!
      products(limit: Int = 10, offset: Int = 0): [Product!]!
    }

    """
    Search result
    """
    type ProductSearchResult {
      totalProducts: Int!
      filteredProducts: Int!
      filters: [LoadedFilter!]!
      products(limit: Int = 10, offset: Int = 0): [Product!]!
    }

    type AssortmentSearchResult {
      totalAssortments: Int!
      assortments(limit: Int = 10, offset: Int = 0): [Assortment!]!
    }
```

For ex. the last one could be:

```
type AssortmentSearchResult {
      totalCount: Int! @deprecated(reason: "Renamed, use the assortmentCount field")
      assortmentCount: Int!
      assortments(limit: Int = 10, offset: Int = 0): [Assortment!]!
    }
```

packages/core-assortments/db/helpers.js
Mikearaya: @pozylon while i agree making the field names uniform is good i think some of the fields that you outlined above are descriptive and shouldn't be deprecated like  `filteredProducts` while `examinedProducts` can be replaced with `productsCount`

pozylon: Agreed, makes sense

package.json
```diff
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "lint": "eslint packages tests --cache --fix --ext=js --ext=ts",
-    "pretest": "npm run lint --silent && npm run typecheck",
```
pozylon, line 23: why that? we should make sure that the code is linted

pozylon (changes requested):

package.json
Mikearaya: @pozylon typescript fails on locally that why i remove it and forgot, putting it back.

Mikearaya:
@pozylon this is ready

package.json
pozylon: @Mikearaya ping

pozylon (approved):

Mikearaya:
@pozylon :raised_hands:

Issue #302, opened 2021-01-20
Use BigInt for Money.amount


schmidsi:
The JavaScript Number type [has many flaws that we need to properly address](https://www.avioconsulting.com/blog/overcoming-javascript-numeric-precision-issues). Most notably this one:

```js
0.1 + 0.2
// => 0.30000000000000004
```

So I propose to switch internally to BigInt and use [graphql-scalars#BigInt](https://github.com/Urigo/graphql-scalars#bigint) in the schema. Furthermore, I propose to add a `decimals` field to the `Amount` type, which defaults to 2. With that, we are perfectly prepared for ERC20 tokens and other currencies which have more than 2 decimals.

We already decided to use graphql-scalars here: https://github.com/unchainedshop/unchained/issues/256

Further reading:
- https://www.avioconsulting.com/blog/overcoming-javascript-numeric-precision-issues
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
- https://www.npmjs.com/package/@melonproject/token-math

PR #343, opened 2021-03-28
Add aditional scalar #302 #256

Mikearaya:


Mikearaya:
I've found minor glitch of migrating to graphql-scalars 
it parses the input to string and this creates a conflict with product schema which currently is a Number type and throws exception when passed a string.
i see to different solutions for this
1. use graphql-scalars BigInt implementation, update product price schema to string and adjust every implementation that expects a number to expect string a do the conversion when needed.
2. Create our own BigInt scalar type.

Mikearaya:
@schmidsi @pozylon

schmidsi (commented):
Very important to have BigInt in order to support crypto-payments

Mikearaya:
@schmidsi so what is your preferred alternative, from the one I've listed. or do you have another?

schmidsi:
> @schmidsi so what is your preferred alternative, from the one I've listed. or do you have another?

I just wanted to say that this is an important change. Maybe also check the codebase for any internal calculations and switch them to BigInt.

Furthermore, we need to store the data as NumberDecimal into mongodb: https://docs.mongodb.com/manual/core/shell-types/#numberdecimal

Mikearaya:
##Current used timestamp value for a date and updates that will be made
### Generic types
```

type Assortment {
created: Timestamp
updated: Timestamp
}


type Bookmark {
created: Timestamp
}


type DeliveryProvider {
created: Timestamp
updated: Timestamp
deleted: Timestamp
}


type Dispatch {
shipping: Timestamp
earliestDelivery: Timestamp
}


type Filter {
updated: Timestamp
created: Timestamp
}

type Log {
created: Timestamp!
}


type PaymentProvider {
created: Timestamp
updated: Timestamp
deleted: Timestamp
}

type Quotation {
created: Timestamp!
expires: Timestamp
updated: Timestamp
isExpired(referenceDate: Timestamp): Boolean
fullfilled: Timestamp
rejected: Timestamp
}


type SubscriptionPeriod {
start: Timestamp!
end: Timestamp!
}


type Subscription {
created: Timestamp!
expires: Timestamp
updated: Timestamp
isExpired(referenceDate: Timestamp): Boolean
}


type LoginMethodResponse {
tokenExpires: Timestamp!
}


type UserLoginTracker {
timestamp: Timestamp!
}

type UserProfile {
birthday: Timestamp
}



type WarehousingProvider {
created: Timestamp
updated: Timestamp
deleted: Timestamp
}

type Work {
started: Timestamp
finished: Timestamp
created: Timestamp!
updated: Timestamp
deleted: Timestamp
scheduled: Timestamp
}


interface OrderDelivery {
delivered: Timestamp
}


type Order {
created: Timestamp
updated: Timestamp
ordered: Timestamp
confirmed: Timestamp
fullfilled: Timestamp
}


interface OrderPayment {
paid: Timestamp
}


interface Product {
created: Timestamp
updated: Timestamp
published: Timestamp
}


type ProductReview {
created: Timestamp
updated: Timestamp
deleted: Timestamp
}


type ProductReviewVote {
timestamp: Timestamp!
}

```



### Mutations
```
addWork( scheduled: Timestamp ): Work

finishWork( started: Timestamp, finished: Timestamp ): Work!

```
### Input types
````
input UserProfileInput {
birthday: Timestamp
}


input DateFilterInput {
start: Timestamp
end: Timestamp
}
```

Mikearaya:
if string representation for the below types is ok we can represent it as BigInt but if Integer is required to be returned i don't thing we can use BigInt.
```
    input UpdateProductCommercePricingInput {
      amount: Float!
    }

   type GeoPosition {
      latitude: Float!
      longitude: Float!
      altitute: Float
    }

    type Price {
      amount: Float!
    }

    type ProductCatalogPrice {
      amount: Float!
    }
```
Else i think float is the closes we can get to long number safely

pozylon:
> type GeoPosition {
>       latitude: Float!
>       longitude: Float!
>       altitute: Float
>     }

This should definetely stay as Float

pozylon:
For BigInt I think there is still a lot to do to make it work, we could prepare tests for BigInt support by testing Ether using WEI prices:

1,00000000000000000000000000000001 ETH
-> 100000000000000000000000000000001 WEI

Mikearaya:
@pozylon

pozylon (changes requested):
I only tried to wrap my head around the timezone topic and started to click around the controlpanel to see if dates/times make sense that are presented but the controlpanel crashed at my first action (trying to add a new product)

- Please make sure the controlpanel is working appropriately with the new fields

Then I had a look at the BigInt topic and even though you have added tests this cannot work,
for example product-catalog.price.js has a line like:

```
const itemTotal = price.amount * quantity;
```

quantity is Int and price.amount could be BigInt, but nodejs does not support mixing int and bigint in calculations.

More than that core-pricing/pricing-sheet uses various calculations that are not safe for BigInt and those fn are used in all pricing plugins. For example there is extensive usage of the `Math` class which does not work with BigInt at all (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math)

pozylon:
@Mikearaya Rebased this and force-pushed to fix Enrollment conflicts

pozylon:
- Add a general test if birthdays are stored and returned correctly (to master)
- Run the test on this branch too to test dates

Issue #116, opened 2019-11-06
Always create a cart when there is a user


claudiocro:
Currently a cart for a user only gets created when there is logged in user/guest and this user adds a product into cart, else the cart is `null`. At first glance this looks ok. 

We are running into a problem with this logic. We are forced to refetch the cart after the first product has been added, we cannot use the response from updateCart to update our apollo cache, as the cart in the cache is null and after the user adds the first product to cart we will get an `_id` for this new cart. Apollo cache does not update the cart in the cache because they are different. We are forced to refetch some cart queries in order to make everything work.

A solution would be to always create a cart for the user. After the user create user, after login if the cart is null and after the user checks out cart, just recreate a new one.

pozylon:
I see, or maybe create a cart when me.cart would return null (in the query resolver)

schmidsi:
Isn't it an anti-pattern to alter the database on a query?

Issue #332, opened 2021-03-03
Invalid payment provider if user adds discounts and outstanding value is zero


schmidsi:
**Describe the bug**
During checkout the following situation was observed in one of our projects:

- User selects payment via credit card
- User adds a voucher to the cart -> outstanding value is now zero
- Now the credit card payment provider is not valid anymore: You cannot pay for 0 on most payment providers

Currently, this leads to a actual invariant: The selected paymentProvider is not part of the supportedPaymentProviders
```json
{
  "data": {
    "me": {
      "cart": {
        "payment": {
          "_id": "ZDFTomSCrF4agbWJq",
          "provider": {
            "type": "GENERIC",
            "interface": {
              "label": "Datatrans Legacy (https://docs.datatrans.ch/v1.0.1/docs/getting-started-home)"
            }
          }
        },
        "supportedPaymentProviders": [
          {
            "_id": "free",
            "type": "INVOICE",
            "interface": {
              "label": "Invoice (manually)"
            }
          }
        ]
      }
    }
  }
}
```

**Expected behavior**
The engine should switch to a valid payment provider. In this special scenario, maybe invoice and mark it as paid immediately? So my suggestion is to change the the result of the `addCartDiscount` mutation from `OrderDiscount!` to `Order`. That way, we can change the `order.payment.provider` inside of this mutation and a front-end can automatically update.

```graphql
addCartDiscount(
orderId: ID
code: String!
): OrderDiscount!
```

to 

```graphql
addCartDiscount(
orderId: ID
code: String!
): Order!
```

macrozone:
the problem does not only apply to updating discounts. usually the filter provider function is defined in userland, so you can't know what can cause a provider to become invalid.


The best solution is  that cart.payment.provider resolves to null if it becomes invalid. In other words: that resolver should use the filter functions as well.

Alternativly this resolver could return a "default payment provider" (the first or a configured one) when its null. The checkout mutation itself should use the same default payment provider.

This would also solve a problem that i observed when i showcased unchained to someone in the playground:
You need to do multiple mutations until the cart can be checked out: login, add products, set payment provider, set delivery provider, set address. i think setting payment provider and delivery provider should not be mandatory and instead use the above mentioned default provider (the same value that cart.payment.provider returns).

pozylon:
All cart actions trigger an invalidation so this feature is actually already built and has been tested, there needs to be another reason in that specific case that thinks the payment provider is still valid.

[...]

PR #296, opened 2021-01-12
Add Recurring Stale User Removal Job

pozylon:

Issue #236, opened 2020-09-25
[FEATURE REQUEST] Disable initialisation of providers on carts on every instance restart


macrozone:
**Introduction**

unchained iterates over all carts (orders) on start and tries to sanitize the current selected providers.
The idea is that carts always have valid providers, e.g. after one has been removed

**Is your feature request related to a problem? Please describe.**

initialisation of providers calls a filter function that can be used to define which providers are available for a customer. e.g. disable certain payment providers if the user is not considered credit worthy. These checks can be expensive (both computational and money), so they have to bee called as rarely as possible.

Currently the filter function is called very often, e.g. whenever you fetch providers through graphql, so every expensive call has to be cached.

Additionally, unchained re-initialize all open carts whenever an instance starts. That is very problematic, because the list of open carts will often grow and will contain staled carts. As the filter function does not know why it is called, you can not reliably tell whether you should call these expensive functions.

in scaleable cloud infrastructures its even worse as instances might come and go. its in general advisable to not touch the database at all on instance starts. Migrations and background jobs should be done in a specific instance

**Describe the solution you'd like**

I would argue that this mechanism of re-initialize the providers on open carts is not worth the drawbacks, I would argue that any consumer of the current cart properties should deal with the situation where the current selected provider is invalid. For frontends, thats usually not a big deal as you can simply tell the users to select a payment provider. Most checkout user flow will handle this. The second consumer is the checkout mutation. This probably should just throw an error when no payment or delivery provider is selected.

**Describe the design of the solution in detail**

see above.

**Describe alternatives you've considered**

setting a valid payment provider could also be done lazily when calling it and it is invalid. That would be even better than the current solution as providers can become invalid at any time (= they get deleted in the db) 

**Additional context**

pozylon:
@macrozone How would you feel if we just allow you to disable the invalidation on startup for now and then find better ways in #332 ?

pozylon:
Possible with beta

PR #344, opened 2021-03-29
always  Initialize user with a cart #116

Mikearaya:
wrong issue number used on commit message. this is for #116


pozylon (changes requested):
Not so happy about this one:

1. it doesn't feel like mutations is the right abstraction layer to add this function, we should propably have an Orders.ensureCartForUser() function that we can call at various places more in the core of unchained. For login creating methods, the hook where we Orders.migrateCart is kind of the right place, for situations where a cart becomes an order (can happen even when a checkout function throws and also when carts get removed)
2. This could be a breaking behavior for some customers and maybe even limit some use cases, in some situations clients use "createCart" mutation to set an arbitrary orderNumber for example. So I think while it's okay to make it a default to ensure having carts on users, we should provide a "settings" parameter to disable this feature
3. We need a bootup call to ensure cart creation for all currently existing users when this feature is first activated
4. With this feature activated, we will propably have much more "dead" carts because of guest logins creating carts and that is linked to #332. To tackle that, we should remove guests after a while and make sure cart invalidation is optimized (see ticket #332 for that).

I propose to take care of 1, 2 & 3 but then kind of tag the feature "experimental" so it's only the default when some sort of experimental environment variable is set to some flag. That way we already have the feature and we can find solutions together with the needed cleanup and invalidation changes for guests and orders.

pozylon:
@Mikearaya I started to work on a stale user removal plugin a while ago but didn't continue:

https://github.com/unchainedshop/unchained/pull/296

Mikearaya:
@pozylon I have add the changes for 1, 2 & 3 like discussed above now there is a environment variable `ENSURE_USER_HAS_CART` required enable this feature and also a setting parameter  `assignCartForUsers` boolean value that can be passed on engine start up to enable cart creation for existing users in a system (default to false).

packages/api/resolvers/mutations/checkoutCart.js
```diff
@@ -10,9 +10,11 @@ export default async function checkoutCart(
   log('mutation checkoutCart', { orderId, userId });
   const cart = await getCart({ orderId, user, countryContext });
   try {
-    return cart.checkout(transactionContext, {
+    const order = cart.checkout(transactionContext, {
```
pozylon, line 13: discard please, await gone, exceptions not catched

packages/core-products/db/products/helpers.js
```diff
@@ -119,6 +120,12 @@ Products.findProducts = ({
   const selector = {};
   const options = { sort };
 
+  log(
+    `hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh ${JSON.stringify(
```
pozylon, line 124: hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh :)

packages/core-products/db/products/helpers.js
```diff
@@ -822,3 +829,284 @@ Products.helpers({
 
 Products.getLocalizedTexts = (productId, locale) =>
   findLocalizedText(ProductTexts, { productId }, locale);
+
+const X = {
```
pozylon, line 833: rm X

packages/core-users/db/helpers.js
```diff
@@ -317,23 +326,32 @@ Users.createLoginToken = async (user, rawContext) => {
...
+      })
       .toArray();
     return (userArray || []).map((item) => new Users._transform(item)); // eslint-disable-line
   }
 
-  return Users.find(selector, { skip: offset, limit }).fetch();
+  return Users.find(buildFindSelector(query), { skip: offset, limit }).fetch();
+};
+
+Users.count = async (query) => {
```
pozylon, line 345: Doesn't this belong to https://github.com/unchainedshop/unchained/pull/342 ?

pozylon (changes requested):
The first time I have reviewed the PR, you have added `await getCart` to various mutations now this is gone and we have a dedicated Orders.ensureCartForUser function and it looks like that function gets called on checkout and on boot and it has been done in a way that is compatible with the webhooks and subscription logic, but now the function seems not to get called anymore when a login token gets created, cart gets migrated or when a cart gets removed. And then there is code from other branches and temp debugging code checked in so I think maybe you forgot to push the latest changes?

packages/core-products/db/products/helpers.js
Mikearaya: :disappointed: sorry i've forgot

Mikearaya:
@pozylon you must have missed it ` Orders.ensureCartForUser` gets called in `LoginTokenCreated` handler https://github.com/unchainedshop/unchained/blob/default-user-cart/packages/platform/setup-accounts.js#L69

pozylon (changes requested):
Everything looks great but we need integration tests for the ENSURE_USER_HAS_CART feature that checks the edge cases in which a user could have no cart. I'd enable ENSURE_USER_HAS_CART in jest-global-setup.js file

pozylon:
@Mikearaya I have another request, can you extend your new function intializeUsersCart and move Orders.invalidateProviders() from core-orders/orders.js to that new function and also make it configurable, so we can solve https://github.com/unchainedshop/unchained/issues/236#issuecomment-810370355 like I suggested

pozylon:
Going to merge this without new tests 🤟🎸 we can add those later

Issue #357, opened 2021-05-04
[FEATURE REQUEST] Cover Images for Assortments


pozylon:
**Introduction**
Unlike for products, it's currently not possible to upload images for assortments.

**Describe the solution you'd like**
The possibility to upload a single cover image for an assortment through the API and Controlpanel.

PR #365, opened 2021-05-24
Add assortment media

Mikearaya:
Solves #357. basically same functionality available for product media is available here. I know the issue requests for a single page but I don't think it's necessary to add that restriction but let the use case dictate how many images assortments should have. 
even though the meta field will be removed in the coming version, assortment media still support it to comply with API `Media` schema
Major addition is the new collection `assortment_document`.


packages/core-assortments/db/assortment-media/schema.js
```diff
@@ -0,0 +1,61 @@
...
+      authorId: { type: String, required: true },
+      title: String,
+      subtitle: String,
+      ...Schemas.timestampFields,
+    },
+    { requiredByDefault: false }
+  )
+);
+
+Migrations.add({
```
pozylon, line 40: The migration should not be needed for a completely new feature

packages/core-assortments/db/migrations.js
```diff
@@ -0,0 +1,70 @@
...
+              authorId: 1,
+            },
+          }
+        );
+      });
+  },
+});
+
+export default () => {
+  runAssortmentMediaMigrations();
```
pozylon, line 68: what you import here is not migrations, it's index creation? please check this code again

pozylon (approved):
Be careful when moving files (packages/core-assortments/db/collections.js → ...assortments/db/assortments/collections.js) and extra check for imports for core-assortments and if references still work, search for "core-assortments" in the whole codebase.

pozylon:
Sorry, didn't want to approve, please look at my comments first before you merge

Mikearaya:
Ok, I'll remove migration and rename it appropriatly. as far as moving files I've checked for any unintended issues, but I'll double-check again

PR #366, opened 2021-05-25
Order search

Mikearaya:
test search valid  for
```
{ _id, userId,  orderNumber,  status  }
```

### DIFF
--- a/packages/api/resolvers/queries/orders.js
+++ b/packages/api/resolvers/queries/orders.js
@@ -1,7 +1,7 @@
 import { log } from 'meteor/unchained:core-logger';
 import { Orders } from 'meteor/unchained:core-orders';
 
-export default function orders(
+export default async function orders(
   root,
   { limit, offset, includeCarts, queryString },
   { userId }
--- a/packages/api/resolvers/queries/orders.js
+++ b/packages/api/resolvers/queries/orders.js
@@ -3,9 +3,11 @@ import { Orders } from 'meteor/unchained:core-orders';
 
 export default function orders(
   root,
-  { limit, offset, includeCarts },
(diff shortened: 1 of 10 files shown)


packages/api/resolvers/queries/orders.js
```diff
@@ -3,9 +3,11 @@ import { Orders } from 'meteor/unchained:core-orders';
 
 export default function orders(
```
pozylon, line 4: not async but Orders.findOrders is async now

packages/core-orders/db/orders/schema.js
```diff
@@ -47,7 +47,7 @@ Orders.attachSchema(
 );
 
 Migrations.add({
-  version: 20200915.2,
+  version: 20210525.1,
```
pozylon, line 50: You should create a new Migrations.add( and always keep the old one.

But anyways we should get away from the migrations when it comes to index creation, it was inflicting too many errors in the past. Do it like here: https://github.com/unchainedshop/unchained/blob/40570b5bb6a233dc0891b7fba3c64c5aae3b0ef4/packages/core-users/db/schema.js#L261

So it will try to create all indexes and if something goes wrong it just removes all indexes and then tries it again

pozylon (commented):
Apart from index creation all looks good, I didn't check the UI yet though

packages/core-orders/db/orders/schema.js
Mikearaya: got it

Mikearaya:
@pozylon all is set, UI is straight you can take my word :)

pozylon (approved):

PR #368, opened 2021-05-29
Env fix

Mikearaya:
Things adjusted
- provide .env.schema path to dotenv-extended load parameter because it was not picking it up.
- provide errorOnMissing: false instead of `process.env.SUPPRESS_ENV_ERRORS` is not being read and it throws an error for missing config values or requests to put default values for each property in .env.schema
- add UNCHAINED_ prefix for COUNTRY, LANG, and CURRENCY env values because I noticed lang value in env file was getting overwritten by os value and that is not intended.


pozylon:
@Mikearaya 

.env-schema was not thought to be loaded by dotenv-extended as it's part of the root and should show all possible env's used while testing, running controlpanel and engine.

I actually don't want to deactivate errorOnMissing because that's one of the most important features of dotenv-extended. What if we just remove it and build two new .env.schema in examples/minimal and examples/controlpanel? To suppress missing env error just set a default value with .env.defaults where possible.

pozylon (commented):
See comment in conv

Mikearaya:
yes, that would get both things done

Mikearaya:
the control panel only has `.env.schema` definitions that are relevant to it. all the rest are in minimal because that's where they are used

examples/controlpanel/next.config.js
```diff
@@ -1,10 +1,10 @@
-require('dotenv').config();
+require('dotenv-defaults').config({ debug: process.env.DEBUG });
```
pozylon, line 1: please favor dotenv-extended instead of dotenv + dotenv-defaults because we used that package in other projects too: https://github.com/keithmorris/node-dotenv-extended

examples/controlpanel/package.json
```diff
@@ -41,7 +41,7 @@
     "cytoscape": "^3.19.0",
     "cytoscape-dagre": "^2.3.2",
     "date-fns": "^2.21.3",
-    "dotenv": "^8.6.0",
+    "dotenv-defaults": "^2.0.2",
```
pozylon, line 44: same here, dotenv-extended combines both

changelog.md
```diff
@@ -1,4 +1,5 @@
 # vNEXT
+- [core] Ambiguity of where to set environment variable is fixed. now you can set environment related with the control panel at the root directory of `controlpanel` and environemt variables related with the engine under root directory of `minimal`.
```
pozylon, line 2: can you add a new section about breaking changes, because it looks like various env's like LANG, CURRENCY and COUNTRY cannot be used anymore and the prefixed env variables have to be used instead, that is breaking.

pozylon (changes requested):

examples/minimal/node_env.js
```diff
@@ -1,6 +1,6 @@
 require('dotenv-extended').load({
   silent: process.env.SUPPRESS_ENV_ERRORS,
-  errorOnMissing: !process.env.SUPPRESS_ENV_ERRORS,
+  errorOnMissing: process.env.SUPPRESS_ENV_ERRORS,
   errorOnRegex: !process.env.SUPPRESS_ENV_ERRORS,
```
pozylon, line 4: sure errorOnMissing should be anebled when the suppress env flag is sent? That seems wrong

pozylon (changes requested):
All nice and clean despite the suppress thing, after you can merge

Mikearaya:
damn, forgot after test :frowning:

Issue #361, opened 2021-05-05
[FEATURE REQUEST] ProductReview should be filterable


macrozone:
**Introduction**

Product Reviews can't be filtered nor sorted, there is only pagination. 

**Is your feature request related to a problem? Please describe.**

Customers want to be able to sort and filter reviews based on different properties

**Describe the solution you'd like**

the resolver Query.productReviews should have sorting and filtering

**Describe the design of the solution in detail**

In our case the customer wants to filter and sort by:

- rating
- some properties of the product

**Describe alternatives you've considered**

Alternative is to load all reviews (remove limit and offset) and sort in-memory.

**Additional context**

I think all list resolvers should have consistent ordering and filtering. With just mongodb, this can become tricky, as you can't sort or filter by relations (at least without using the aggregation pipeline). I think relational data stores have a big advantage here. E.g. with prisma and nexus, graphql queries with nested sorting and filtering is trivial. I think migrating to an ORM or similar that solves these kind of problems would be a big step forward.

pozylon:
Very reasonable to sort and filter by rating, author or text search on title and comment

pozylon:
Will be part of beta9

PR #374, opened 2021-06-14
Product review filter

Mikearaya:
#361 product review support search with (title & review) and sorting with any valid property of product review.


pozylon:
I have to think about the sort option you have added here

pozylon:
LGTM

PR #376, opened 2021-06-23
Matomo tracker update

Mikearaya:
currently, the Product category is not available (in publicare context) unless we use tags here after.

### DIFF
--- a/packages/core-events/plugins/matomo-tracker.js
+++ b/packages/core-events/plugins/matomo-tracker.js
@@ -2,14 +2,13 @@ import fetch from 'isomorphic-unfetch';
 import { encode } from 'querystring';
 import { subscribe } from '../director';
 
-const extractOrderParameters = (orderPosition) => {
+const extractOrderParameters = (order) => {
   const orderOptions = {};
   orderOptions.idgoal = 0;
-  orderOptions.ec_id = orderPosition.orderId;
+  orderOptions.ec_id = order._id;
 
   orderOptions.ec_items = JSON.stringify(
-    orderPosition
-      .order()
+    order
       .items()
       .map((i) => [
         `${i.productId} SKU`,
@@ -21,8 +20,8 @@ const extractOrderParameters = (orderPosition) => {
   );
   // eslint-disable-next-line no-underscore-dangle
   orderOptions._ects = new Date().getTime();
-  const pricing = orderPosition.pricing();
-  orderOptions.revenue = orderPosition.order().pricing().total().amount;
+  const pricing = order.pricing();
+  orderOptions.revenue = order.pricing().total().amount;
   orderOptions.ec_tx = pricing.taxSum();
   orderOptions.ec_dt = pricing.discountSum();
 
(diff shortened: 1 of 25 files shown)


packages/core-events/plugins/matomo-tracker.js
```diff
@@ -2,28 +2,31 @@ import fetch from 'isomorphic-unfetch';
 import { encode } from 'querystring';
 import { subscribe } from '../director';
 
-const extractOrderParameters = (orderPosition) => {
+const parseCurrency = (amount) => parseFloat(amount / 100).toFixed(2);
```
pozylon, line 5: Why you do a parseFloat here? Please use accounting package if you do divide and toFixed. Else we will have problems with the float handling of js (http://openexchangerates.github.io/accounting.js/)

pozylon (commented):
Still not tracking orders and carts correctly... 

> Tracking Cart Updates in Matomo (Optional)
> 
> To track cart additions and removals, your cart system will need to send the details for every item that remains in the cart after a user adds or removes an item, including those already submitted from prior Add to Cart clicks. This is important to collect accurate information for the abandoned cart feature within Matomo. The data should be included within a “push”, which is the function that enables sending of structured data to Matomo via JavaScript. You will also need to supply the total value of all items in the cart. so the two required elements for cart tracking are:
> 
> A push of 'addEcommerceItem' for each product currently in the cart, including the name and SKU at a minimum.
> A final 'trackEcommerceCartUpdate' push with the cart total passed as a parameter.

Can't see trackEcommerceCartUpdate anywhere or removeEcommerceItem or clearEcommerceCart...

This will not be very stable like this I guess...

Mikearaya:
@pozylon  see https://matomo.org/docs/ecommerce-analytics/#how-to-set-up-a-new-ecommerce-website-in-matomo

`addEcommerceItem` is supposed to handle all those operations

Mikearaya:
@pozylon 
To track cart additions and removals, your cart system will need to send the details for every item that remains in the cart after a user adds or removes an item, including those already submitted from prior Add to Cart clicks. This is important to collect accurate information for the abandoned cart feature within Matomo. The data should be included within a “push”, which is the function that enables sending of structured data to Matomo via JavaScript. You will also need to supply the total value of all items in the cart. so the two required elements for cart tracking are:

    A push of 'addEcommerceItem' for each product currently in the cart, including the name and SKU at a minimum.
    A final 'trackEcommerceCartUpdate' push with the cart total passed as a parameter.

Mikearaya:
@pozylon  ok, noted. i used parseFloat to provide correct value as int/float

pozylon:
@Mikearaya Can you please stale this PR for the moment and copy the Matomo Tracker Plugin to the Publicare Repository so we have everything in one place. We can re-add it back to the engine once it works as expected.

pozylon:
Introduced async_hooks async context through meteor 2.3 (node 14), will be the base for reading context data in a future matomo tracker version

Mikearaya:
@pozylon nothing is being registered in the context yet right?

PR #377, opened 2021-07-19
Extend matomo

Mikearaya:
I think the hybrid approach we are implementing now is here to stay because search (`assortmentSearch` and `productSearch`) breaks the context unless used with rawCollection and some useful functions like setDocumentTitle are not provided for pageView tracking through HTTP API.
In addition, you can retrieve the context if available through the third parameter of `tranform` function passed to the MatomoTracker instance


packages/core-events/plugins/matomo-tracker.js
```diff
@@ -51,30 +51,28 @@ const MatomoTracker = (siteId, siteUrl, subscribeTo, options = {}) => {
...
+
+    matmoOptions = options?.transform
+      ? options?.transform(subscribeTo, matmoOptions, context) ?? {}
+      : matmoOptions;
+
     await fetch(
       `${siteUrl}/matomo.php?idsite=${siteId}&rec=1&action_name=${
         actionMap[subscribeTo]
-      }&${encode(orderOptions)}${extraOptions}`
+      }&${encode(matmoOptions)}`
```
pozylon, line 68: typo matmoOptions

packages/core-events/plugins/matomo-tracker.js
```diff
@@ -51,30 +51,28 @@ const MatomoTracker = (siteId, siteUrl, subscribeTo, options = {}) => {
...
     );
   });
 };
 
-export default MatomoTracker;
```
pozylon, line 80: I'd additionally export MatomoTracker and not only the convenience method

pozylon (approved):

PR #380, opened 2021-07-26
Remove meta resolver

Mikearaya:
each payment provider should implement a type for PaymentCredential meta to be removed DataTrans, Strip, etc... because each expect different fields


pozylon:
Documentation looks very good and else the changes are fine, which cases where we have arbitrary payload data that we have to generally make available to the frontend, still exist now? @Mikearaya

Mikearaya:
@pozylon yes, projects need to define their own explicit resolvers now and no meta field will be available for the listed types. meta will be available at a resolver definition point

Mikearaya:
anything you think should be added to this?

Mikearaya:
@pozylon

pozylon:
Meta Usage Today by the big shops:
- Order.meta
- Assortment.meta
- OrderDelivery.meta

pozylon:
I fixed some things with controlpanel and removed the order meta props, tests all running through

pozylon (approved):

PR #385, opened 2021-08-13
Add new assortment link breadcrumb resolver that saves n+1 calls to A…

pozylon:
…ssortments.findOne


pozylon:
Biggest issue: 

- createAssortmentLink and createAssortmentProduct do not check if the provided childAssortmentId or productId exist
- the helpers addLink and addProduct (+ addFilter, ...) on assortment call createAssortmentLink/createAssortmentProduct respectively, also without checking if provided data results in failed links
- the mutation that call addLink, addProduct and addFilter ARE CHECKING THAT IT'S VALID
- the bulk importer DOES NOT check if data is valid

PR #389, opened 2021-09-03
Datatrans v2

pozylon:
This plugin is developed from the ground up with the following new features compared to the old now legacy Datatrans Plugin:

- 100% Typescript
- Uses the new Datatrans JSON based API's
- Supports marketplace comissions
- Supports the whole bandwidth of currently supported acquirers of Datatrans (https://www.datatrans.ch/de/features/zahlungsmittel/) including Apple Pay and and Android Pay
- Supports secure fields flows
- Defers settlements wherever possible and settles the transaction only when checkout is actually called and the transaction is valid. Before there were edge cases where people were beeing charged with some amount but the Unchained Order failed to transition away from the cart state
- Cancels transactions when Unchained finds the transaction unsuited
- Supports multiple stored credentials per payment type, for ex. multiple credit cards


Mikearaya (commented):
@pozylon The first thing I looked for was docs update and changelog. can you add few statements either in changelog or docs

pozylon:
Holy, sure that is missing completely, thanks

pozylon:
@Mikearaya I did some extended documentation about how to use the datatrans plugin, what do you think: https://github.com/unchainedshop/unchained/blob/5730ef9e22b6496e022c9c201a71c649e2072b71/docs/source/plugins/datatrans.md

Mikearaya:
@pozylon it's really informative. I took the liberty to edit while reading

PR #390, opened 2021-09-13
Draft: Jm/unchained mongodb package

pozylon:


packages/core-events/director/configureEventDirector.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      if (!_registeredEvents.has(eventName))
+        throw new Error(`Event with ${eventName} is not registered`);
+
+      _adapter.publish(eventName, {
+        payload: { ...data },
+        context: extractedContext,
+      });
+
+      Events.insert({
```
pozylon, line 65: await

pozylon: + we should propably move this code to some "insertEvent" fn

packages/core/core.js
```diff
@@ -1,35 +1,38 @@
...
+
+export const initCore = async ({
   modules = {},
   migrationRepository,
   ...otherComponents
 } = {}) => {
   const moduleOptions = {
     migrationRepository,
   };
-  configureLogger(modules.logger, moduleOptions);
```
pozylon, line 32: I think the db connection should open explicitly here and not at "import { db } from 'meteor/unchained:core-mongodb';" so it should be more of

```
import { connect } from 'meteor/unchained:core-mongodb';

initCor = () {
...
   const db = connect(MONGO_URL, mongoDBOptions);

   configureBla({db, ...rest});
}
```

pozylon (commented):
I think we are totally on the right path here and it's a great base for further discussion, let's get into details once you're back from vacation @joelmeiller

PR #391, opened 2021-09-14
Replaced 'Control Panel' with 'Admin UI' etc.

aleksandra-unchained:


pozylon:
Nice! thanks @aleksandra-unchained

PR #398, opened 2021-11-09
Event expiry

Mikearaya:


pozylon:
Thanks @Mikearaya, @joelmeiller will include the mongodb index expiry date in his refactoring

PR #418, opened 2022-02-14
Cryptopay Payment & Pricing Plugin

OpenCoreCH:
Payment & pricing plugins for the crypto payment gateway. For the pricing plugin, the main business logic is implemented by the generic `ProductPriceRateConversion` plugin which consumes the rates in the database with the `products.prices.rates` module. The rates in the database are produced by the web hook calls.
The same pattern can be used for the Coinbase / ECB plugins such that they only feed the rates in the DB, but do not contain any business logic.

The payment plugin supports payments for products that have a crypto price or for products that only have a fiat price (in this case, the new rate plugin is used for conversion, so the plugin can also be used in conjunction with different rate providers).


pozylon (approved):
Very nice! can you please:

- add some documentation on how to use the payment plugins in conjunction with crypto pay
- add some documentation on how to use the rate providing functions
- and write a changelog entry stating what core features have been added (rate stuff)

Issue #423, opened 2022-02-17
[FEATURE REQUEST] Postfinance Checkout Flex Plugin


pozylon:
https://www.postfinance.ch/de/unternehmen/produkte/e-commerce/postfinance-checkout-flex.html?msclkid=3c0713621e5f152b80a0716c29e549be&utm_source=bing&utm_medium=cpc&utm_campaign=(a_gk)(p_checkout)(l_de)(n_gsn)(s_checkout)(t_brand)&utm_term=postfinance%20e%20payment&utm_content=E-Payment%20Lösung

- [x] Test Environments
- [x] API Documentation found
- [x] Extended tests same like for Datatrans V2
- [x] Support for Simulation and Live Mode
- [x] Support for all supported payment methods "Unterstützte Akzeptanzpartner und Zahlungsarten"
- [x] Support for alias storage for cards ("save payment details")
- [x] Support for Deferred Settlement / Asynchronous Capturing and cancelation of reservations (very important for customer case!)

PR #424, opened 2022-02-24
PostFinance Checkout Plugin

OpenCoreCH:
Implements new PostFinance Checkout Plugin (#423) and adds support to Unchained for deferred settlement / cancellations / refunds.


pozylon:
😍

pozylon (approved):

PR #425, opened 2022-02-25
Add GridFS File Adapter and refactor files and file-upload packages

pozylon:


pozylon:
Missing: removeFiles in zombie killer

packages/mongodb/src/db/initDb.ts
```diff
@@ -1,27 +1,21 @@
...
-  // Use connect method to connect to the server
-  log('Initialise MongoDB database', { color: LogTextColor.Crimson });
-  await client.connect();
-  log('Client successfully connected', { color: LogTextColor.Crimson });
-  const db = client.db(dbName);
-  log('Dabase successfully initialised', { color: LogTextColor.Crimson });
-  return db
-  */
+const initDbMeteor = async (): Promise<Db> => {
   return MongoInternals.defaultRemoteCollectionDriver().mongo.db as unknown as Db;
```
OpenCoreCH, line 16: Only a small detail, but the `as unknown as Db` seems superfluous to me.
`.mongo.db` is of type `MongoDb` and there should not be a type error without the hints

packages/file-upload/plugins/gridfs/gridfs-adapter.ts
```diff
@@ -0,0 +1,136 @@
...
+  version: '1.0',
+
+  ...FileAdapter,
+
+  async createSignedURL(directoryName = '', fileName) {
+    const expiryDate = getExpiryDate();
+    const _id = hash(`${directoryName}-${fileName}-${expiryDate.getTime()}`);
+    const signature = sign(directoryName, _id, expiryDate.getTime());
+
+    const putURL = `${ROOT_URL}gridfs/${_id}/${directoryName}/${fileName}?e=${expiryDate.getTime()}&s=${signature}`;
```
OpenCoreCH, line 56: Does the `ROOT_URL` always contain a trailing slash? For instance, the datatrans plugin has `${ROOT_URL}/payment/datatrans/webhook` as a default value in its configuration, so I assumed this would be usually without a trailing slash

packages/file-upload/plugins/minio-webhook.js
```diff
@@ -0,0 +1,49 @@
...
+      const { Records = [], EventName } = req.body;
+      if (EventName === 's3:ObjectCreated:Put' && isAuthorized(headers)) {
+        const [{ s3 }] = Records;
+        const { object } = s3;
+        const { size, contentType: type } = object;
+        const [fileId] = object.key.split('.');
+        const { services } = req.unchainedContext;
+        await services.files.linkFile({ fileId, type, size }, req.unchainedContext);
+        res.writeHead(200);
+        res.end();
```
OpenCoreCH, line 32: I think you should return here to avoid that the `writeHead(404)` is also called on the response object (like in https://docs.meteor.com/packages/webapp.html).
Not sure what the behavior is when `writeHead` is called after the response is closed, probably results in an error

OpenCoreCH:
@pozylon Very nice! I added some minor comments and I think you missed the `await` for one `initDb` call in `delivery-index.test.ts`, looks otherwise good to me.

packages/file-upload/plugins/gridfs/gridfs-adapter.ts
pozylon: yeah, not always! better have one too much than one too less

PR #430, opened 2022-03-10
New mutation rejectOrder

### LINKED ISSUE #428: Order Rejection Feature "rejectOrder"
Should put the Order into a final new "REJECTED" status and cancels reservetions and/or refunds payments.

OPEN -> PENDING

PENDING -> CONFIRMED/REJECTED
CONFIRMED -> FULLFILLED

FULLFILLED ->
REJECTED ->

- [x] Generate an E-Mail with a new Template ("Sorry shit hit the fan, you won't get the article")

OpenCoreCH:
Resolves #428


packages/core-orders/src/module/configureOrdersModule-processing.ts
```diff
@@ -248,6 +248,43 @@ export const configureOrderModuleProcessing = ({
...
+        {
+          paymentContext: params.paymentContext,
+          deliveryContext: params.deliveryContext,
+        },
+        requestContext,
+      );
+      const orderPayment = await modules.orders.payments.findOrderPayment({
+        orderPaymentId: order.paymentId,
+      });
+      await modules.payment.paymentProviders.cancel(
```
pozylon, line 278: Does this also refund through Postfinance when the payment has already been paid?

pozylon (approved):
Approved but it should refund if Payment already paid, not sure if does that. It's possible an OrderPayment is paid but Order still PENDING. Could be the case if payment done with Bitcoin and order rejected by Clerk Dude in Orderly Case.

packages/core-orders/src/module/configureOrdersModule-processing.ts
OpenCoreCH: Now it does :)

OpenCoreCH:
> Approved but it should refund if Payment already paid, not sure if does that. It's possible an OrderPayment is paid but Order still PENDING. Could be the case if payment done with Bitcoin and order rejected by Clerk Dude in Orderly Case.

Yes, `cancel` is called on the payment provider of the order payment (now also with `refund` passed in the transaction context, s.t. that providers like Postfinance do not only try voiding, but also refunding), even when the order is PENDING.

PR #449, opened 2022-04-25
Add main user events #419

### LINKED ISSUE #419: Add User Events
Currently, updating a user profile does not emit an event but could be very beneficial for various cases where we have to send data to a CRM, so we should also add events to the users package. @Mikearaya can you look into this and implement it on master?

Mikearaya:
Closes #419


packages/core-users/src/module/configureUsersModule.ts
```diff
@@ -12,6 +13,22 @@ import { FileDirector } from 'meteor/unchained:file-upload';
...
+const USER_EVENTS = [
+  'USER_UPDATE',
+  'USER_UPDATE_PROFILE',
+  'USER_ADD_ROLES',
+  'USER_UPDATE_ROLE',
+  'USER_UPDATE_TAGS',
+  'USER_UPDATE_AVATAR',
+  'USER_UPDATE_HEARTBEAT',
+  'USER_UPDATE_BILLING_ADDRESS',
+  'USER_UPDATE_LAST_CONTACT',
```
pozylon, line 25: Looks good to me, i'd rename "cleanConfidentialData" to "removeConfindentialServiceHashes" to make it a little bit more clear what get's claned, but that's it.

pozylon (approved):

packages/core-users/src/module/configureUsersModule.ts
Mikearaya: I made it general in case we wanted to remove other data, but this name is descriptive for the current purpose, coming up.

PR #451, opened 2022-04-27
Update user profile custom data store

### LINKED ISSUE #405: Mutation.updateUserProfile replaces existing unrelated data with new
**Describe the bug**
if you call `updateUserProfile` 2 times with `profile.customFields` set to different values it will replace the value set on the first call with the value given on the second call.

**To Reproduce**
call
```
updateUserProfile(userId: 'x', profile: {
customFields: {
phoneNumbers: [{number: '0912887733', type: 'mobile'}]
}
})


updateUserProfile(userId: 'x', profile: {
customFields: {
addresses: [{city 'x', country: 'y'}]
}
})

```
in this case, `phoneNumbers` will be replaced

**Expected behavior**
it should not replace values set to different keys other than the ones provided on the actual call.

Mikearaya:
Closes #405

Instead of storing custom data in `customFields` it's stored in `meta` field plus it top-level field and not under profile  for consistency with other types and `customFields` is obsolete


packages/@unchainedshop/types/user.ts
```diff
@@ -52,6 +51,7 @@ export type User = {
   services: any;
   tags?: Array<string>;
   username?: string;
+  meta: any;
```
pozylon, line 54: Sure this can't be null?

docs/source/config/users.md
```diff
@@ -14,3 +14,18 @@ const options = {
   }
 };
 ```
+### Storing additional values
+
+The standard user data type stores most of the information a user could have. But if you want to store additional information in user, you can do so by extending `UserProfileInput` that is passed to the `Mutation.updateUserProfile` to store the additional fields you want :
+
+```
+extend input UserProfileInput {
```
pozylon, line 22: I'm not sure if this is the right approach, if you look at how meta is changed in other mutations, it looks like this:

```
input UpdateProductInput {
      tags: [String!]
      sequence: Int
      meta: JSON
    }
```

So the input is not extended but there is a meta field in the input, only the type to read from is extended.

packages/core-users/src/module/configureUsersModule.ts
```diff
@@ -198,13 +198,32 @@ export const configureUsersModule = async ({
 
     updateProfile: async (_id, profile, userId) => {
```
pozylon, line 199: The customFields was a half-baken solution which was not clear what to expect from.

I agree with you that if you use "extension" of input parameters, the actual update profile should only update fields that have been set but leave the others as is.

But like I stated above, that would not be consistent: We should propably transition to a way where we set an arbitrary "meta" field which could be anything JSON-stringifiable. Then we only have 1 variable to set and because it's arbitrary it's clear for the API user that if he sets meta, he overwrites what was meta before.

So I'd keep the simpler method as before, if meta is sent, overwrite meta.

pozylon (changes requested):

pozylon:
After thinking about it we should propably only:

1. rename customFields field from input type to meta to make it consistent with other update mutations that touch custom data
2. remove the customFields field from the output type
3. store meta data of the user not under profile but under root

I'd keep the current way of replacing the whole field with new data as is. We cannot know for what kind of additional data this feature is used so we cannot assume they only want to send an object that "patches" on the db.

packages/@unchainedshop/types/user.ts
```diff
@@ -52,6 +51,7 @@ export type User = {
   services: any;
   tags?: Array<string>;
   username?: string;
+  meta?: any;
```
pozylon, line 54: Isn't this on the profile? the meta field is on the profile in the input type?

docs/source/config/users.md
```diff
@@ -14,3 +14,12 @@ const options = {
   }
 };
 ```
+### Storing additional values
+
+The standard user data type stores most of the information a user could have. But if you want to store additional information in user, you can do so by using the `meta` field on  `UserProfileInput` that is passed to the `Mutation.updateUserProfile`:
```
pozylon, line 19: Now that meta is on root, it does not make sense to provide meta in UserProfileInput, it should be another toplevel:

The standard user data type stores most of the information a user could have. But if you want to store additional information in user, you can do so by using the `meta` field ~~on  `UserProfileInput`~~ that is passed to the `Mutation.updateUserProfile`: 

->

The standard user data type stores most of the information a user could have. But if you want to store additional information in user, you can do so by using the `meta` field as a property on `Mutation.updateUserProfile`

packages/@unchainedshop/types/user.ts
```diff
@@ -90,7 +90,7 @@ export type UsersModule = {
   updateInitialPassword: (user: User, initialPassword: boolean) => Promise<void>;
   updateLastBillingAddress: (_id: string, lastAddress: Address, userId: string) => Promise<User>;
   updateLastContact: (_id: string, lastContact: Contact, userId: string) => Promise<User>;
-  updateProfile: (_id: string, profile: UserProfile, userId: string) => Promise<User>;
+  updateProfile: (_id: string, profile: UserProfile & { meta?: any }, userId: string) => Promise<User>;
```
pozylon, line 93: Doesn't feel right, can we have updateProfile: (_id: string, { profile?: UserProfile, meta?: any }, userId: string) => Promise<User>

pozylon (changes requested):
I think we can simplify it a little more and make it more concise.

packages/@unchainedshop/types/user.ts
pozylon: Or maybe another "updateMeta"

Mikearaya: @pozylon why not both? i'll implement both

Mikearaya:
@pozylon it is now

pozylon:
CI tests ran manually on machine -> all good, false negative

PR #456, opened 2022-05-09
Fix optional assortment selector check #455

### LINKED ISSUE #455: Query.assortments is not filtering root assortments only
*The bug**
By default `query.assortments` accepts a `includeLeaves` parameter which is supposed to return when leaf node when it's truth. currently, it seems like this filter is not working and is returning all assortments (leaf and root) all the time instead

Mikearaya:
this expression is always false
```
  if (!assortmentSelector && !includeLeaves) {
    selector.isRoot = true;
  }
```
because `!{}` is false and so is `!{key: value}`

**Updated**

 ```
  if (!Object.keys(assortmentSelector || { }).length && !includeLeaves) {
    selector.isRoot = true;
  }
```

Closes #455

### DIFF
--- a/packages/core-assortments/src/module/configureAssortmentsModule.ts
+++ b/packages/core-assortments/src/module/configureAssortmentsModule.ts
@@ -55,7 +55,7 @@ const buildFindSelector = ({
     }
   }
 
-  if (!assortmentSelector && !includeLeaves) {
+  if (!Object.keys(assortmentSelector).length && !includeLeaves) {
     selector.isRoot = true;
   }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Mikearaya** [comment]
@pozylon if a user sets `assortmentSelector` based on some conditions and in one condition it is passed `{}` then it will be a problem.

eg
on publicare migration, there was a line like this
```
    context.modules.assortments.search.findFilteredAssortments({
      limit: 0,
      offset: 0,
      assortmentIds,
      assortmentSelector: selector,
      sort: { index: 1 },
    });
```

pozylon:
> @pozylon if a user sets `assortmentSelector` based on some conditions and in one condition it is passed `{}` then it will be a problem.
> 
> eg on publicare migration, there was a line like this
> 
> ```
>     context.modules.assortments.search.findFilteredAssortments({
>       limit: 0,
>       offset: 0,
>       assortmentIds,
>       assortmentSelector: selector,
>       sort: { index: 1 },
>     });
> ```

I just checked that line and it works perfect because it's a custom selector and it's important that it returns exactly the assortments that is provided by the assortmentIds

PR #458, opened 2022-05-13
Make product and assortments tag filter case insensitive #437

### LINKED ISSUE #437: [FEATURE REQUEST] Tag search should not be case sensitive
When searching with tags the current implementation is search input tag should match exactly including case, but that should not be the case.

Mikearaya:
Closes #437

### DIFF
--- a/packages/core-assortments/src/module/configureAssortmentsModule.ts
+++ b/packages/core-assortments/src/module/configureAssortmentsModule.ts
@@ -22,6 +22,15 @@ import { configureAssortmentTextsModule } from './configureAssortmentTextsModule
 import { configureAssortmentMediaModule } from './configureAssortmentMediaModule';
 import { makeAssortmentBreadcrumbsBuilder } from '../utils/breadcrumbs/makeAssortmentBreadcrumbsBuilder';
 
+const convertTagsToLowerCaseIfExists = ({ tags = [], ...assortment }: Assortment): Assortment => {
+  const normalized = {
+    ...assortment,
+  };
+  if (tags && Array.isArray(tags)) normalized.tags = tags.map((tag) => tag.toLowerCase());
+
+  return normalized;
+};
+
 const ASSORTMENT_EVENTS = [
   'ASSORTMENT_CREATE',
   'ASSORTMENT_REMOVE',
@@ -50,7 +59,7 @@ const buildFindSelector = ({
 
   if (tags) {
     if (Array.isArray(tags)) {
-      selector.tags = { $all: tags.map((tag) => new RegExp(`^${tag}$`, 'i')) };
+      selector.tags = { $all: tags.map((tag) => tag.toLowerCase()) };
     } else {
       selector.tags = tags;
     }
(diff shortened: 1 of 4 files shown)


packages/core-assortments/src/module/configureAssortmentsModule.ts
```diff
@@ -265,6 +265,7 @@ export const configureAssortmentsModule = async ({
         skip: offset,
         limit,
         sort: { sequence: 1 },
+        collation: { strength: 2, locale: 'en' },
```
pozylon, line 268: What does this do?

packages/core-assortments/src/module/configureAssortmentsModule.ts
```diff
@@ -49,7 +49,7 @@ const buildFindSelector = ({
 
   if (tags) {
     if (Array.isArray(tags)) {
-      selector.tags = { $all: tags };
+      selector.tags = { $all: tags.map((tag) => new RegExp(`^${tag}$`, 'i')) };
```
pozylon, line 52: @Mikearaya Isn't it better to use

https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/toLowerCase

instead of a regex? like this it's using the regex search for every tags based lookup which might affect the performance.

pozylon (changes requested):
I wonder if it's the right approach and i'm a bit skeptical if this is rock solid.

As an alternative we could consider transforming all tags lowercase on insert/update into db, when querying and add a small migration that converts all existing entries to lowercase. Then we can have very performant lookups and don't need any advanced indexes or collation, which might slow down the whole thing. And what happens if I store a non english term in tags like "öhse", does that work fine with collation?

What do you think? @Mikearaya

pozylon:
@Mikearaya I'm much more comfortable without regex and i think it's worth the change

Mikearaya:
tags are now stored in lowercase and search terms also converted to lowercase

pozylon:
I'm sorry @Mikearaya but I'm not really happy with the quality of this PR and the overall implementation, there is many things that I don't like at the moment:

1. convertTagsToLowerCaseIfExists is two times same code (not DRY)
2. convertTagsToLowerCaseIfExists is run from a very low level update call, so it's applied to a field that could contain MongoDB $set , $addToSet $pull.
3. what about all the upper case tags in the db that already exist? convertTagsToLowerCaseIfExists is only run on products/assortments that are changed, but should be a migration instead that runs 1 time over all existing tags.
4. there is not just tags on products and assortments, there is many more tags in unchained, for example media, assortment links etc.

I suggest solving it like this:

- Completely dropping this PR, start from scratch, we will do this with Unchained 1.1, so this is not a priority for now.
- Write a migration that migrates the tags in the db for all entities that currently can have tags

**For the update/insert case:**

You can never enforce MongoDB to validate that the entry is lowercase on a low level because mongodb update operations can be written in 1000 ways, that means you have to:
- do it on the complete other side of the road (= closest to API surface -> graphql & bulk-import) 
- communicate this lowercase tag concept to the user/developer so he knows he has to go lowercase with his own tags

An elegant way to solve the graphql part in Mutation is is by either using a graphql scalar type "LowerCaseString" that validates the tag format in all the input fields or by using a directive that you apply to input fields "tags" o.e.: https://www.apollographql.com/docs/apollo-server/v2/schema/creating-directives#example-uppercasing-strings


**When querying:**

See above, use directives or graphql scalar types to communicate that the tags you supply to a Query

PR #459, opened 2022-05-13
Event filters

### LINKED ISSUE #400: Enable multiple event type filter for events
**Introduction**
currently, events query only allows filter for a single event type

**Describe the solution you'd like**
events should accept filter for multiple event types

**Describe the design of the solution in detail**
change events query to accept an array of event types and adjust the engine accordingly

**Additional context**

Mikearaya:
Closes #400  
as an added bonus I've added sorting for `query.workQueue` which can accept any field value

Also, there were many inconsistencies between count and list filter queries in which count methods were not accepting some of the parameters of the list.  eg. `query.filters` accepted `queryString` but `Query.filtersCount` did not. this PR solves this for all queries.

finally, a single source of typing is used for API and module list queries


Mikearaya:
@pozylon yes it is ready, forgot to switch...i will rebase after

pozylon (changes requested):
Looks good but I have one remark:

Can we @deprecated the "selectTypes" and also call it "types" like in events?

workQueue(
        limit: Int = 10
        offset: Int = 0
        status: [WorkStatus!]
        **selectTypes: [WorkType!]**
        created: DateFilterInput
        queryString: String
        sort: [WorkQueueSortOptionInput!]
      ): [Work!]!

pozylon:
@Mikearaya + tests fail currently, that needs to be fixed too. master currently passes

packages/api/src/schema/inputTypes.js
```diff
@@ -19,6 +19,15 @@ export default [
       key: String!
       value: SortDirection!
     }
+    input WorkQueueSortOptionInput {
+      key: String!
+      value: SortDirection!
+    }
+
+    input EventSortOptionInput {
```
pozylon, line 27: one remark:

why duplicate and not refactor it to have a single input type "SortOptionInput"? instead of ProductReviewSortOptionInput, WorkQueueSortOptionInput, EventSortOptionInput that all look the same?

Now you implement another concept with the events rather than using one of the existing concepts of orderBy in searchAssortments or sort in reviews (i'd prefer the review way).

packages/api/src/schema/query.js
```diff
@@ -382,12 +398,19 @@ export default [
...
       Get all emitted events
       """
-      events(type: String, limit: Int = 10, offset: Int = 0, queryString: String): [Event!]!
+      events(
+        types: [String!]
+        limit: Int = 10
+        offset: Int = 0
+        queryString: String
+        created: DateTime
+        sort: [SortOptionInput]
```
pozylon, line 407: [SortOptionInput] -> [undefined] okay

pozylon (approved):

pozylon (approved):
A very very good quality PR with a clean and tidy refactoring! Very good job @Mikearaya !

Just look at the comments there was two topics I found a bit weird (created not beeing a range in a query and that isArray check of types). You can fix that in an upstream pr so you don't have to rebase all the other PR's.

packages/core-events/src/module/configureEventsModule.ts
```diff
@@ -1,20 +1,17 @@
...
-  type?: string;
-  queryString?: string;
-};
-const buildFindSelector = ({ type, queryString }: FindQuery) => {
-  const selector: { type?: string; $text?: any } = {};
+const buildFindSelector = ({ types, queryString, created }: EventQuery) => {
+  const selector: { type?: any; $text?: any; created?: any } = {};
 
-  if (type) selector.type = type;
+  if (types && Array.isArray(types)) selector.type = { $in: types };
```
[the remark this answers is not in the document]
Mikearaya: @pozylon just  making it extra safe.

PR #460, opened 2022-05-13
Workqueue count api

### LINKED ISSUE #401: Expose total work queue count  API
like all the other list collections There need to be an API for the work queue that tells how much work is currently saved in the work queue. it's necessary for pagination.

Mikearaya:
Closes #401 

**Note: this PR is an extension of #459 and shares on some values so that should be merged before this**

### DIFF
--- a/packages/api/src/resolvers/queries/worker/workQueueCount.ts
+++ b/packages/api/src/resolvers/queries/worker/workQueueCount.ts
@@ -1,0 +1,33 @@
+import { log } from 'meteor/unchained:logger';
+import { Root, Context } from '@unchainedshop/types/api';
+import { WorkStatus } from '@unchainedshop/types/worker';
+
+export default async function workQueueCount(
+  root: Root,
+  {
+    status,
+    selectTypes,
+    created,
+    queryString,
+  }: {
+    queryString?: string;
+    status?: Array<WorkStatus>;
+    selectTypes?: Array<string>;
+    created?: {
+      start: Date;
+      end: Date;
+    };
+  },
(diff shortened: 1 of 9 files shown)


packages/api/src/resolvers/queries/worker/workQueueCount.ts
```diff
@@ -0,0 +1,33 @@
...
+  root: Root,
+  {
+    status,
+    selectTypes,
+    created,
+    queryString,
+  }: {
+    queryString?: string;
+    status?: Array<WorkStatus>;
+    selectTypes?: Array<string>;
```
pozylon, line 15: selectTypes is deprecated, the new workQueueCount should only support the non deprecated "types"

packages/api/src/schema/inputTypes.js
```diff
@@ -20,6 +20,11 @@ export default [
       value: SortDirection!
     }
 
+    input EventSortOptionInput {
```
pozylon, line 23: We now have a general sort input I think

packages/api/src/schema/query.js
```diff
@@ -381,6 +381,16 @@ export default [
         types: [WorkType!]
       ): [Work!]!
 
+      """
+      Return total number of workers filtered the provided arguments
+      """
+      workQueueCount(
+        status: [WorkStatus!]
+        selectTypes: [WorkType!]
```
pozylon, line 389: Like above: only support types

pozylon (changes requested):

PR #479, opened 2022-06-29
Fix cryptopay plugin context query

Mikearaya:
`cryptoAddresses` is  stored as an individual object in context `{contractAddresses: {currency, address }}` but the queries are creating a selector like the fields are stored individually inside context `{currency, address }`


Mikearaya:
https://github.com/unchainedshop/unchained/blob/master/packages/core-payment/plugins/cryptopay.ts#L178

Mikearaya:
@pozylon

pozylon:
@Mikearaya please rebase then merge

pozylon (approved):

PR #493, opened 2022-08-11
Fix types

### LINKED ISSUE #491: Fix type error noise when building engine
There are a lot of typescript inconsistencies and it's creating a lot of noise when building the engine

pozylon:
Fixes #491


Mikearaya (commented):
Now it's much better, there the type errors have reduced very much. but I still get these logs when I run it on watch mode, is it related to the UnchainedApi context field update you mentioned on the issue? 
I think `index.d.ts` is missing a few exports

Screenshot from 2022-08-15 22-09-55
Screenshot from 2022-08-15 22-10-03
Screenshot from 2022-08-15 22-10-20
Screenshot from 2022-08-15 22-10-52

Mikearaya:
however, there are no import errors detected when opening a file in vscode, which was the case previously, so looks good to me until the API context is adjusted.

pozylon:
Weird that you still get any ts issues, it runs without here, can you run it again?

Mikearaya:
i don't see any commit, What did you change, now there are no errors :smiley:

Mikearaya (approved):

pozylon:
The problem is that there is an inexplicit order, for ex. @unchainedshop/types should be compiled first, then the rest. Will solve that later.

PR #506, opened 2022-09-19
Convert tags to lower case

### LINKED ISSUE #437: [FEATURE REQUEST] Tag search should not be case sensitive

Mikearaya:
Closes #437 

- Added custom LocaleCaseString scalar that does the conversion on GraphQL API level
- Convert tags to lower case for all entities in bulk import 
- Added migration for all collections tag fields that convert them to lowercase.

while working on this, I noticed a minor issue with findFilters typing so I have fixed that too


pozylon (commented):
In general:
- Looks good conceptually!
- User also has tags but no migration?
- I don't like the utils function "convertTagsToLowerCase" because it is a util that is only needed just 1 time during migrations and adding it to a package as util function somehow does not feel right.

packages/core-users/src/module/addMigrations.ts
```diff
@@ -1,7 +1,27 @@
...
+    if (count % 500 === 0) {
+      bulk.execute();
+      bulk = collection.initializeUnorderedBulkOp();
+      count = 0;
+    }
+  }
+  if (count > 0) bulk.execute();
+};
+
 export default function addMigrations(repository: MigrationRepository<Migration>) {
```
Mikearaya, line 25: @pozylon User has a migration too

pozylon:
@Mikearaya We need an admin ui branch to fix the breaking changes with the tags, can you prepare that too so we can merge both of that together when ready?

pozylon (approved):

Mikearaya:
@pozylon prepared we can merge it when this is merged https://github.com/xecutors/unchained-adminui/pull/301

PR #521, opened 2022-11-28
Add unit test setup for packages #483

### LINKED ISSUE #483: Fix all unit tests
After migration from meteor to npm, the existing package unit tests don't work and there is a concept missing:

- Use jest for everything
- Add test setup to every package
- Add at least 1 passing test to every package
- Migrate exiting mocha/chai tests

Mikearaya:
Closes #483


packages/core-assortments/jest.config.js
```diff
@@ -1,5 +1,14 @@
 /** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
 export default {
-  preset: 'ts-jest',
+  preset: '@shelf/jest-mongodb',
```
pozylon, line 3: Unit tests should not test the db, that is beeing done with the integration tests, only test pure functions

pozylon (commented):
It's important to define what we test with unit tests, and it should be mainly pure functions, an example excerpt from code:

```
const divideTreeByLevels = (
  array: Tree<string>,
  level = 0,
): Array<{ level: number; items: Array<string> }> => {
  const currentLevel: Array<string> = array.reduce((acc, item) => {
    if (typeof item === 'object') {
      return acc;
    }
    return [...acc, item];
  }, []) as Array<string>;

  const nextLevels = array.reduce((acc, item) => {
    if (typeof item === 'object') {
      return [...acc, ...divideTreeByLevels(item, level + 1)];
    }
    return acc;
  }, []) as Array<{ level: number; items: Array<string> }>;

  return [currentLevel.length && { level, items: currentLevel }, ...nextLevels].filter(Boolean);
};
```

perfectly testable function because it's pure!


```
update: async (assortmentFilterId, doc) => {
      const selector = generateDbFilterById(assortmentFilterId);
      const modifier = { $set: doc };
      await AssortmentFilters.updateOne(selector, modifier);
      return AssortmentFilters.findOne(selector, {});
    },
```
practically untestable because it would require stubbing AssortmentFilters, if then it is tested, nothing is left so it's not beneficial at all to test that code with unit tests.

So the idea is when using unit tests, only test the actually testable functions which are pure and for later, make code testable by applying techniques from "Refactoring" http://martinfowler.com/books/refactoring.html

packages/core-assortments/jest.config.js
pozylon: A third example of code is the file build-paths.ts. It contains only pure functions but takes functions as an argument. Such a code is unit testable as well because we can supply a stub function that usually involves db calls:

```
const paths = await walkUpFromAssortment({
  resolveAssortmentLink: (assortmentId) => {
    if (assortmentId === 'subsubchild') {
      return {
        assortmentId,
        parentIds: ['subchild'],
      };
    }
    if (assortmentId === 'subchild') {
      return {
        childAssortmentId: assortmentId,
        parentIds: ['root'],
      };
    }
    if (assortmentId === 'root') {
      return {
        childAssortmentId: assortmentId,
        parentIds: [],
      };
    }
    return null;
  },
  assortmentId: 'subsubchild',
});

expect(paths).toEqual([{ links: ['subsubchild', 'subchild', 'root'] }]);
```

pozylon: @Mikearaya

pozylon (changes requested):
Looks quite good already! Can you make sure the tests are run in non-watch mode when doing "npm run test" and introduce a special test:watch ? (consistency with root package.json)

+ can you run all tests of all packages when doing npm run test:packages in the root?

pozylon:
@Mikearaya i just merged master into this pr because there were some conflicts, and i fixed some regressions that you introduced, make sure to pull

pozylon:
Ok, i merged master into this PR (please don't rebase at this state with so many changes in between and forks), currently there is still some "watch" in some packages, i tried "npm run test -ws" + due to the master mege some tests fail now (some improved selectors)

PR #522, opened 2022-11-28
 Improve Deletion of Products #477 

### LINKED ISSUE #477: Deleting products does not delete assortment product links

Mikearaya:
Closes #477

### DIFF
--- a/packages/api/src/resolvers/mutations/products/removeProduct.ts
+++ b/packages/api/src/resolvers/mutations/products/removeProduct.ts
@@ -32,7 +32,7 @@ export default async function removeProduct(
   const openQuotation = await modules.quotations.openQuotationWithProduct({ productId });
   if (openQuotation)
     throw new ProductLinkedToQuotationError({ productId, quotationId: openQuotation?._id });
-  const openEnrollment = await modules.enrollments.openEnrollmentsWithProduct({ productId });
+  const openEnrollment = await modules.enrollments.openEnrollmentWithProduct({ productId });
   if (openEnrollment)
     throw new ProductLinkedToEnrollmentError({ productId, enrollmentId: openEnrollment?._id });
 
--- a/packages/api/src/resolvers/mutations/products/removeProduct.ts
+++ b/packages/api/src/resolvers/mutations/products/removeProduct.ts
@@ -6,6 +6,7 @@ import {
   ProductLinkedToActiveVariationError,
   ProductLinkedToActiveBundleError,
   ProductLinkedToEnrollmentError,
+  ProductLinkedToQuotationError,
 } from '../../../errors';
 
 export default async function removeProduct(
@@ -30,7 +31,7 @@ export default async function removeProduct(
     if (e?.message === 'ProductLinkedToEnrollmentError')
       throw new ProductLinkedToEnrollmentError({ productId });
     if (e?.message === 'ProductLinkedToQuotationError')
-      throw new ProductLinkedToEnrollmentError({ productId });
+      throw new ProductLinkedToQuotationError({ productId });
 
     throw e;
   }
--- a/packages/api/src/resolvers/mutations/products/removeProduct.ts
+++ b/packages/api/src/resolvers/mutations/products/removeProduct.ts
@@ -22,11 +22,12 @@ export default async function removeProduct(
   if (!(await modules.products.productExists({ productId })))
     throw new ProductNotFoundError({ productId });
 
-  const activeLink = await modules.products.firstActiveProductLink(productId);
-  if (activeLink?.bundles?.length)
-    throw new ProductLinkedToActiveBundleError({ productId, bundleId: activeLink?._id });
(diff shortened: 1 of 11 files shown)


packages/core-products/src/module/configureProductsModule.ts
```diff
@@ -15,6 +15,10 @@ import {
   buildSortOptions,
 } from '@unchainedshop/utils';
 import { SortDirection, SortOption } from '@unchainedshop/types/api';
+
+import { configureEnrollmentsModule } from '@unchainedshop/core-enrollments';
+import { configureQuotationsModule } from '@unchainedshop/core-quotations';
```
pozylon, line 20: this is not allowed by design (core module can't import from other core modules), find another way please

pozylon: see removeProductService.ts for reference

pozylon (changes requested):

pozylon:
@Mikearaya i just merged master into this pr because there were some conflicts, make sure to pull

packages/core-products/src/module/configureProductsModule.ts
```diff
@@ -364,10 +365,29 @@ export const configureProductsModule = async ({
 
     delete: async (productId) => {
       const product = await Products.findOne(generateDbFilterById(productId), {});
-
       if (product.status !== InternalProductStatus.DRAFT) {
         throw new Error(`Invalid status', ${product.status}`);
       }
+      const activeLinks = await Products.findOne(
```
pozylon, line 371: Some checks are here...

packages/core-products/src/service/removeProductService.ts
```diff
@@ -1,16 +1,27 @@
...
+export const removeProductService: RemoveProductService = async ({ productId }, unchainedApi) => {
+  const { modules } = unchainedApi;
   const product = await modules.products.findProduct({ productId });
   switch (product.status) {
     case ProductStatus.ACTIVE:
       await modules.products.unpublish(product);
     // falls through
     case null:
     case ProductStatus.DRAFT:
+      const hasActiveQuotations = await modules.quotations.quotationExistsForProduct({ productId });
```
pozylon, line 14: and some are here...

packages/api/src/resolvers/mutations/products/removeProduct.ts
```diff
@@ -14,8 +21,20 @@ export default async function removeProduct(
 
   if (!(await modules.products.productExists({ productId })))
```
pozylon, line 22: and some checks are here...

pozylon (commented):
Okay while technically this implementation works, i'm having a bit of a problem with the exception concept here and I think the denial checks should be on the top level, why?

1. The removeProduct service is called by the removeProduct mutation and is also called by the bulk import. The bulk import should be more brute force and basically just work so there should not be any errors. But currently both the callee "removeProductService" and the callee of the callee "module.products.delete" throw errors. The only situation when removing a product failed from the point of view of the bulk import was when the product was already removed and in that situation, it's just getting logged to bulk import work and does not have any negative effect on the whole bulk import.

2. Routing the string errors from inside the services and core to the mutation and rethrowing with typed error adds bloat and even removes crucial details that are important, actually when receiving an error in the UI I should be able to parse the conflicting bundle when removing a simple product from a bundle.


Suggestions:
1. Do all checks directly in the mutation but add the functions to check for those issues 
2. Don't just check for `enrollmentExistsForProduct` but use something like `openEnrollmentsWithProduct` and `bundlesWithProduct` then throw the first blocking productId so the error is thrown like this: throw new ProductLinkedToActiveVariationError({ productId, configurableProductId });

pozylon:
reasoning why a bulk import should be able to remove a product that is part of an active bundle is that the bundle removal event could be part of the same batch but comes after.

packages/core-products/src/module/configureProductsModule.ts
```diff
@@ -361,10 +362,29 @@ export const configureProductsModule = async ({
...
+        _id: activeLink?._id,
+        variations:
+          activeLink?.proxy?.assignments?.filter(
+            ({ productId: variantProductId }) => variantProductId === productId,
+          ) || [],
+        bundles:
+          activeLink?.bundleItems?.filter(
+            ({ productId: bundledProductId }) => bundledProductId === productId,
+          ) || [],
+      };
```
pozylon, line 384: Strongly suggest to make 2 functions, this is mixing stuff that has no relation to each other

pozylon (changes requested):

Mikearaya:
I merged bundle and proxy query into one to save DB requests and that is the same reason I queried for the single product instead of a list. (firstActiveProductProxy)

pozylon:
@Mikearaya ok i see that, i think if it were a query that would have to be something to think of but mutations are usually triggered from users and we can afford 300ms

Mikearaya:
@pozylon  It's ready though, i've separated them

PR #524, opened 2022-11-30
Move to Core Types

pozylon:
Big Refactor:
- Remove createdBy
- Remove updatedBy
- Remove deletedBy
- Remove authorId except for product reviews
- Only use Context in API, Core modules should work independent from the API layer environment

Why?
- Removes almost 1000 lines of bloat code
- Opens up the whole framework to core service introduction the end goal of unchained software architecture is below


pozylon:
Removes almost 1000 lines of code

PR #533, opened 2022-12-22
Add admin UI platform configuration option

Mikearaya:


packages/api/src/context.ts
```diff
@@ -13,7 +13,12 @@ export const setCurrentContextResolver = (newContext: UnchainedContextResolver)
 };
 
 export const createContextResolver =
-  (unchainedAPI: UnchainedCore, roles: any, version: string): UnchainedContextResolver =>
+  (
+    unchainedAPI: UnchainedCore,
+    roles: any,
+    version: string,
+    adminUiConfig: AdminUiConfig,
```
pozylon, line 20: I think it should be optional type

packages/api/src/schema/types/shop.js
```diff
@@ -4,13 +4,24 @@ export default [
       href: String
       title: String
     }
+
+    type CustomEntityInterface {
```
pozylon, line 8: the type name should be something like AdminUiConfigCustomEntityInterface (prefixed with admin ui), else "CustomEntityInterface" is too general

packages/api/src/schema/types/shop.js
```diff
@@ -4,13 +4,24 @@ export default [
...
+    }
+
     type Shop @cacheControl(maxAge: 180) {
       _id: ID!
       language: Language
       country: Country
       version: String
       userRoles: [String!]!
       externalLinks: [Link]!
+      adminUiConfig: AdminUiConfig
```
pozylon, line 24: You're doing extra work to ensure that adminUiConfig is set to an object with an empty array but in GraphQL you define adminUiConfig as optional, i think it should be AdminUiConfig! here

pozylon (changes requested):
Looks quite good, only a few minor things

PR #536, opened 2022-12-30
Nodenext

pozylon:
A customer wants to run Unchained in Lambda

In order to use Unchained Engine in serverless environments like Amazon Lambda, we need to adhere to default Node.js 16+ ESM rules.

It basically means that imports without explicit file endings are not supported anymore and this PR changes that so that node doesn't have to be run with "-es-module-specifier-resolution node"

In Node18 the flag is already experimental and in Node19 it's been dropped completely so we anyways have to go this way:

https://nodejs.org/docs/latest-v18.x/api/esm.html#customizing-esm-specifier-resolution-algorithm
https://nodejs.org/docs/latest-v19.x/api/esm.html#customizing-esm-specifier-resolution-algorithm


Mikearaya (approved):
Looks done to me

PR #549, opened 2023-02-22
Add oauth2 

Mikearaya:


packages/core-accountsjs/src/director/Oauth2Adapter.ts
```diff
@@ -0,0 +1,40 @@
...
+      isActive: () => {
+        return true;
+      },
+      getAuthorizationCode: async (authorizationCode: string) => {
+        return null;
+      },
+      getAccountData: async (userAuthorizationToken: any) => {
+        return null;
+      },
+      isTokenValid: async (token) => {
```
pozylon, line 25: How would you suggest using isTokenValid and when? How does this influence the login with the user, the session token?

packages/platform/src/setup/setupAccounts.ts
```diff
@@ -71,6 +71,115 @@ export const setupAccounts = (unchainedAPI: UnchainedCore) => {
...
+
+      const userOAuthInfo = await oauth2Service.getAccountData(userAuthorizationToken);
+      console.log(userOAuthInfo);
+
+      if (!userOAuthInfo || !userOAuthInfo?.email) {
+        throw new Error('OAuth authentication failed');
+      }
+
+      try {
+        const user = await unchainedAPI.modules.users.findUser({
```
pozylon, line 104: I thought a lot about this. We shouldn't go for an email match, that could be different users and also users could get disconnected if email is changed on provider side. 

When signing in with an oAuth provider, we should get the unique userId from the provider and check if there is already a user that has that id in "user.services.oauth.PROVIDER". If there is a match, that's the user we are looking for and there we should store the current authorizationToken (and eventually refreshToken).

If there is no match then we should try to register the user with the mapped registration data, email and username. If that doesn't work because the email or username is already used by some other user, cancel the registration (in that case we show something like that to the user: "The E-Mail address or username returned from OAUTH_PROVIDER_XYZ already matches a user in our system, please log in with that user first, you can link it with OAUTH_PROVIDER_XYZ throught the account page after)."

Then let's look at the the situation when somebody created a user with email mike@mike.com and password or mike with WebAuthn before. In those cases the user should be able to "link" his social account on an account page, so we need some mutations for that ("linkWithOAuth" "unlinkFromOAuth") or something like that.


Look at other pages like "dev.to" for example:
Bildschirm­foto 2023-03-07 um 08 10 38

pozylon (changes requested):
Okay i think to have a bullet-proof implementation of oAuth, we need to think about all edge cases:

- What if username/email already taken?
- How can I link existing users?
- When we receive a refresh token, we should use it to keep having access to the provider data, so how do we refresh the access token regularly?
- What kind of oAuth2 Flow do we support, this has to be documented

Mikearaya:
@pozylon isn't email a unique identifier of a particular user? So if a user tries to log in with a email address that is already taken then they will be logged in automatically and no error should be thrown.
Regarding using the refresh token flow, I don't think it's applicable in our case and we only should rely on unchained session expiry, else we will have two sources of truth of a user session duration.

I agree with implementation of linking accounts to get extra information from each one of them. It is basically skiping the services implementation and requesting user data from each service and storing them. Other than data retrieval I don't think there is much use of linking accounts for administration purpose, sharing is the responsibility of a storefront, right?

pozylon:
> @pozylon isn't email a unique identifier of a particular user? So if a user tries to log in with a email address that is already taken then they will be logged in automatically and no error should be thrown. 

This is a security risk. If a user decides to go with email+password+totp because he thinks it's security relevant and somebody hacks his linked in account that guy can also take over the totp-secured account that way.

> Regarding using the refresh token flow, I don't think it's applicable in our case and we only should rely on unchained session expiry, else we will have two sources of truth of a user session duration.

Okay, I agree with that, so the refresh token is only used so we can do authorized requests to the provider for an infinite time. Which is something we might want for some oAuth providers. That still means we need a process to refresh those tokens, so maybe a work queue item OAUTH_REFRESH_AUTHORIZATION_CODES or something.

pozylon:
> I agree with implementation of linking accounts to get extra information from each one of them. It is basically skiping the services implementation and requesting user data from each service and storing them. Other than data retrieval I don't think there is much use of linking accounts for administration purpose, sharing is the responsibility of a storefront, right?

Not quite, I mean if you link an existing email+password user to linkedin, you still have to store the userId and the authorization key on that user's oauth services object because we might need to have access to that backend. For example if we add Github as oauth provider, we could add a custom query to Unchained that uses the authorization key internally to grab a list of repositories from github and return that.

pozylon:
Linking also means we don't copy over the profile data because we are not creating a new user.

Mikearaya:
Linking and unlinking function is added at the service level. refresh token function is also added. however, logging in and linking oauth flow refresh token is unnecessary because we are getting a fresh `access_token` on each request. a refresh token is useful when you constantly retrieve data from an OAuth provider and don't want to ask the user to log in each time. for example, Google's access_token expiry period is 1 minute so we have to refresh the token for no useful reason every minute.
with that said I don't think implementing a worker that refreshes all tokens is necessary but we should provide the interface that will perform that in the plugins so when that use case comes adding a worker that suits the need will be simple for the developer.

**Use the oauth branch of admin UI to test all the functionalities under account page**

Mikearaya:
@pozylon `refresh_token` and other data returned from authentication are stored in DB.  The rest of consistency fix have been made

pozylon (changes requested):
**Conflicting naming:**

Mutation.loginWithOAuth
Mutation.linkOAuthProvider
Mutation.unlinkOAuthProvider
ShopInfo.oauthProviders
User.linkedOAuthProviders

**Can we go:** (Accounts & Providers)

Mutation.loginWithOAuth
Mutation.linkOAuthAccount
Mutation.unlinkOAuthAccount
User.oAuthAccounts: [OAuthAccount!]!

ShopInfo.oAuthProviders: [OAuthProvider!]!


And OAuthAccount should not contain any profile data that is not normalized. That doesn't make a lot of sense to me because we don't know anything about the oAuth provider and the data that we are interested in could dramatically be different so leave this up to the customer to extend the type OAuthAccount.

I propose the following structure for OAuthAccount:

```
type UserOauthData {
       _id: String!
       provider: String!
       authorizationCode: String
     }
```

replacing

```
type UserOauthData {
       authorizationCode: String
       email: String!
       lastName: String
       avatarUrl: String
       firstName: String
       address: String
       gender: String
       birthDate: Date
       phoneNumber: String
       displayName: String
       city: String
       countryCode: String
       postalCode: String
       regionCode: String
       company: String
     }

     type OauthLink {
       provider: String!
       data: [UserOauthData!]!
     }
```

pozylon:
@Mikearaya

I'm not challenging the getAccountData, I'm just saying the GraphQL Type `UserOauthData` is a bad idea because the only thing that is needed to create a user is this:
```
await modules.accounts.createUser({
  username: 'my-super-funny-username',
});
```

So the absolute minimum getAccountData implementation has to return a username field and nothing more, and even if you don't have a username in the oAuth provider, you have the _id which you can then use for the username of the user you're creating:

return { username: `${id}@linkedin` }

pozylon:
@Mikearaya i don't get it, can you please explain me like i'm 5? 😹 why do you need a graphql type that returns an e-mail address of an oauth provider when you don't know if the oauth provider stores e-mails at all?

Mikearaya:
**@pozylon ** what oauth provider does not use email?

PR #552, opened 2023-03-18
Add missing plugin overviews

Mikearaya:


pozylon (changes requested):
Bildschirm­foto 2023-03-22 um 09 54 28

-> Consistency: "Unchained Admin UI" instead of "Admin UI"

We copied the docs initially from https://www.apollographql.com/docs/
They have a "Home" part and once you select Engine or Admin UI you see completely different side nav with the details.

Is this doable too instead of having just a link to one of the sections on the left when clicking admin ui?

pozylon:
http://localhost:8000/admin-ui/currency/ -> Don't use the word "Ethereum" because Unchained could support Solana, Cardano etc.

Always stay blockchain-neutral and write something like: If you add a crypto currency, you can optionally configure a contract address that can be used to distinguish tokens on a blockchain. A cryptocurrency payment plugin (for ex. cryptopay) can then optionally read that information and knows which currency is actually meant.

pozylon:
When you write the docs for users, products and assortments, make sure to add a bug issue for every bug you encounter in admin ui so you can fix those directly on master on that repository, no matter if it's just very small glitches or critical issues. For example I had issues doing variations and bundling stuff.


And also get specific about the product types, there is a lot of questions around that when I discussed this with Joel from Bob Pay, he currently builts a shop based on Unchained Engine for that company as a reference implementation of their service.

pozylon:
Admin UI stuff looks good, we could merge that already and then continue with the finetuning in another PR!

pozylon (approved):

pozylon:
I'm merging this and add a more specific issue about the changes needed

pozylon:
@Mikearaya follow up: https://github.com/unchainedshop/unchained/issues/557

PR #563, opened 2023-06-04
Add meta on bokkmark schema

Mikearaya:
Includes bookmark extension that will enable projects to store and use additional bookmark information using meta fields

- Added meta on bookmarks db schema
- extend `mutation.createBookmark` to accept additional optional `meta` parameter
- extend BookmarkModule to include `deleteByUserIdAndMeta` util for performing bulk delete operation based on meta.

This PR is generic and doesn't implement the exact requirement we encountered for the favorite list on a project. Still, it will enable us to extend the functionality to the specific condition per project bases.

### DIFF
--- a/packages/api/src/resolvers/mutations/bookmarks/createBookmark.ts
+++ b/packages/api/src/resolvers/mutations/bookmarks/createBookmark.ts
@@ -4,7 +4,7 @@ import { BookmarkAlreadyExistsError, InvalidIdError, ProductNotFoundError } from
 
 export default async function createBookmark(
   root: Root,
-  { productId, userId }: { productId: string; userId: string },
+  { productId, userId, meta }: { productId: string; userId: string; meta?: any },
   { modules, userId: currenctUserId }: Context,
 ) {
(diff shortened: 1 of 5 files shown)


pozylon:
I like it, just wondering if we should accept updating the meta property too? I think it would be consistent in Unchained Engine.

Mikearaya:
@pozylon I thought of implementing that too but I since bookmark core allows updating the bookmark, I think we should leave the api implementation to the end developer to implement per project cases.
What do you think?

pozylon:
@Mikearaya I checked and we have both situations in unchained: For example

- createEnrollment takes a meta and also updateEnrollment takes a meta
- createCart does not take a meta but updateCart does
- createProduct does not take a meta but updateProduct does
- addProductReviewVote takes a meta but there is no update mutation

So I'm fine with your proposal, let's leave the meta stuff up to the project, poor man's update could also be removing a bookmark for a product and creating it again with different meta props.

pozylon (approved):

PR #564, opened 2023-06-11
Bookmark fix

Mikearaya:
While working with Bookmark on a project I encountered this issues and fixed them on this PR

- `mutation.bookmark` didn't accept `meta` while it's possible to do the same thing with `mutation.createBookmark` I think it would be great if it accepts too
- **BUG** update operations in bookmark module emitted `BOOKMARK_UPDATED` event but the registered bookmark module event is `BOOKMARKS_UPDATE` .
 

other than the above issues current bookmark module is flexible and stable enough to use it for different application :)

### DIFF
--- a/docs/source/advanced-config/events.md
+++ b/docs/source/advanced-config/events.md
@@ -101,7 +101,8 @@ Below are events tracked under each module under the box:
 
 | Event name      | Emitted when...         | `payload`                |
 | :-------------- | :---------------------- | :----------------------- |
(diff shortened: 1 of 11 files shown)


pozylon:
I've refactored this a little: bookmark mutation only allows as long as there is not multiple bookmarks of a product, else it fails and throws an error.

createBookmark and removeBookmark are kept like done by you mike. the idea should be: if you use meta, don't use bookmark

PR #573, opened 2023-07-31
Impersonator

Mikearaya:
updated impersonation implementation that will let granted users to impersonate a non-admin user account and switch back to their own account without the need to log out and log back in.

### DIFF
--- a/packages/api/src/resolvers/mutations/accounts/stopImpersonation.ts
+++ b/packages/api/src/resolvers/mutations/accounts/stopImpersonation.ts
@@ -10,21 +10,21 @@ export default async function stopImpersonation(root: Root, _, context: Context)
   const impersonatedUser = await modules.users.findUserById(userId);
 
   if (!impersonatedUser) throw new UserNotFoundError({ userId });
-  const impersonator = impersonatedUser?.services?.resume?.loginTokens.find(
+  const impersonatedUserToken = impersonatedUser?.services?.resume?.loginTokens.find(
     ({ hashedToken, impersonatorId }) =>
       hashedToken === modules.accounts.createHashLoginToken(loginToken) && impersonatorId,
   );
 
-  if (!impersonator) throw new Error('Current session is not being impersonated');
+  if (!impersonatedUserToken) throw new Error('Current session is not being impersonated');
 
   await modules.accounts.logout({ loginToken }, context);
 
   const accountsServer = modules.accounts.getAccountsServer();
 
-  const { token } = await accountsServer.loginWithUser(impersonator.impersonatorId);
+  const { token } = await accountsServer.loginWithUser(impersonatedUserToken.impersonatorId);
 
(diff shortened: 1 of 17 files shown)


pozylon:
It was a bit of a missunderstanding, i rebased on mast and added a commit with my thinkings. For me it was important that we have a clear ACL that says: admins can use impersonate, non-admins can't.

packages/api/src/resolvers/mutations/accounts/stopImpersonation.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export default async function stopImpersonation(root: Root, _, context: Context) {
+  const { userId, modules, loginToken } = context;
+
+  log(`mutation stopImpersonation for ${userId}`);
+
+  const impersonatedUser = await modules.users.findUserById(userId);
+
+  if (!impersonatedUser) throw new UserNotFoundError({ userId });
+  const impersonator = impersonatedUser?.services?.resume?.loginTokens.find(
```
pozylon, line 13: Remark: This naming is missleading, one could think the object holds the user object of the impersonator but it actually holds the loginToken of the impersonated user.

packages/api/src/schema/types/user.ts
```diff
@@ -153,6 +153,7 @@ export default [
       allowedActions: [RoleAction!]!
       tokens: [Token!]!
       oAuthAccounts: [OAuthAccount!]!
+      impersonator: User
```
pozylon, line 156: Remark: The impersonator is actually not something that is unique to one user but actually based on the current session / login token used. Consider using a root level Query.impersonator instead of putting it on the actual user. Similar as Query.me

packages/core-accountsjs/src/module/configureAccountsModule.ts
```diff
@@ -100,15 +101,33 @@ export const configureAccountsModule = async ({
       };
     },
 
-    createImpersonationToken: async (userId, rawContext) => {
+    createImpersonationToken: async (userId, rawContext: Context) => {
       // TODO: rawContext does not contain user and locale date anymore
       // following the type but the code still depends on it
-      const context = evaluateContext(filterContext(rawContext));
+      const { modules } = rawContext;
```
pozylon, line 107: Why not use the evaluateContext(filterContext thing? If it's not needed you could remove it everywhere it's used currently to clean it up.

pozylon (changes requested):

packages/core-accountsjs/src/module/configureAccountsModule.ts
Mikearaya: `modules` will be filtered after `evaluateContext(filterContext` operation but we actually only need access to modules, not a normalized context and we cant remove the dependency of this utility functions because they are used to normalize the final token information stored in DB

PR #575, opened 2023-08-07
Add missing  actions

### LINKED ISSUE #570: [FEATURE REQUEST] Add a validation step on mutation resolvers
It would greatly enhance the functionality to include a validation step for user input before executing the actual mutation resolution process. While most of the validation can be performed using plugins, these are typically triggered after a record is created, which may not always be desirable.

To address this, I propose the implementation of a validation layer that can be extended on a per-project basis, similar to modules and services. This layer would run prior to each resolver and would incorporate a configured validation step. The following code snippet demonstrates a possible implementation:

```
startPlatform({
  validators: {
    createUser: {
      someValidation: () => {},
      anotherValidation: () => {}
    }
  }
})
```

By utilizing this approach, we can ensure that all mutation resolvers undergo a validation process, offering enhanced control and flexibility in handling user input.

Mikearaya:
Closes #570 

Added the following actions and link to the corresponding resolver

```
'logout',
logoutAllSessions',
'loginAsGuest',
'loginWithPassword',
'loginWithWebAuthn', 
'loginWithOAuth',
'verifyEmail',
'useWebAuthn', for createWebAuthnCredentialCreationOptions & createWebAuthnCredentialRequestOptions
 because they will probably require the same access control rule
'pageView',
'createUser',
'forgotPassword',
'resetPassword',
'changePassword',
'heartbeat',
```


pozylon:
The rationale behind is that it's then easier to customize for users in their own projects, right? Or why do those anonymous methods need the actions?

Mikearaya:
@pozylon yes, no current behaviour will be affected/changed by this

PR #577, opened 2023-08-25
Move *_TEXT_UPDATE event upsertLocalizedText helper

### LINKED ISSUE #576: Can't subscribe to product upsertLocalizedText operation
Currently, we have two specific methods for modifying product texts: `updateTexts` and `upsertLocalizedText`. However, only the `updateTexts` method triggers the `PRODUCT_UPDATE_TEXTS` event. Unfortunately, there isn't any event associated with the `upsertLocalizedText` method, which makes it impossible to create a subscription for it.

Interestingly, this discrepancy isn't present in other modules where the `upsertLocalizedText` method is employed, as it does trigger an event there.

It would be highly beneficial if the `upsertLocalizedText` function within the products module could also generate an event that we could subscribe to. This would enhance our ability to monitor changes effectively and respond accordingly.

Mikearaya:
Closes #576


pozylon:
Can you please list the mutations that will trigger that event now following the callstack, to check against?

packages/core-assortments/src/module/configureAssortmentMediaModule.ts
```diff
@@ -75,7 +75,12 @@ export const configureAssortmentMediaModule = async ({
       },
     );
 
-    return AssortmentMediaTexts.findOne(selector, {});
+    const mediaTexts = await AssortmentMediaTexts.findOne(selector, {});
```
pozylon, line 78: That is not the way:
1. Extra fetch per locales, no good
3. When updating multiple locales for ex. 3 locales, it will add 3 extra fetches + emit the UPDATE TEXT event 3 times instead of once

pozylon (changes requested):
Important: The way this has been built will lead to many events duplicate because upsertLocalizedText runs per locale, we should look again into the update heuristics

Question: I like moving it to upsertLocalizedText but what do you think if we change the payload structure and emit the event:
1. per locale (for example in the ASSORTMENT_UPDATE_TEXTS case: instead of  `{ assortmentId: string, assortmentTexts: [] }` save  `{ assortmentId: string, text: {} }`) and text containing the locale
2. only emit when the actual update call actually altered a db object and not every time it is called, mongodb's update operation returns various properties to check that.

Mikearaya:
@pozylon 
>     1. per locale (for example in the ASSORTMENT_UPDATE_TEXTS case: instead of  `{ assortmentId: string, assortmentTexts: [] }` save  `{ assortmentId: string, text: {} }`) and text containing the locale

I think it more intuitive this way and by moving the event in upsertLocalizedText the all *_TEXT_UPDATE events payload is a single text object instead of an array like it was previously.

>     2. only emit when the actual update call actually altered a db object and not every time it is called, mongodb's update operation returns various properties to check that.

with how things are wired currently it's not possible to determine if there was an actual update or if nothing changed by calling update because this will always be the value returned by the operation no matter what.  with the exception `upsertedId` being set on the insert operation.
```{
  acknowledged: true,
  modifiedCount: 1,
  upsertedId: null,
  upsertedCount: 0,
  matchedCount: 1
}
```

Mikearaya:
its because of the `updated` field value change with current timestamp that renders an update even when no data is changed too

pozylon:
Yeah, i see the constraints and I'm not happy with either solution. I think this concept needs a complete rethink, i'll invest some time

pozylon:
Okay so I kept most of your changes but I removed the multi-update changes again (first text, then updated date). After thinking about it again I'd rather accept text updates for text that has not changed instead of having a hacky multi-update process. By doing that, I switched to mongodb feature called "returnDocument" with findOneAndUpdate to reduce the roundtrips as a "compromise".

Then I fixed some docs, I changed the event's name to better match what it actually does and I reviewed all the places where the text gets updates so that the create events come after the text updates. That helps fixing the weird situations where you subscribe to a CREATE event and then don't have the initial text.

Lastly, I've removed "public exposure" of the upsertLocalizedText function so it's only called from within the core modules itself reducing the surface of the API a bit

PR #581, opened 2023-10-31
User deletion

### LINKED ISSUE #582: [FEATURE REQUEST] Enable User deletion
General overview

Two ways this can be implemented
- soft-delete
- hard-delete

1. soft-delete:- have some type of flag such as 'DEACTIVATED' or use the existing deleted field. in this case, the user data still exists in the system but should not be visible unless reactivated again.
2. hard-deletion: permanently deletes the user and removes all the footprints a user might have such as bookmarks, orders, subscriptions,s, etc...

Mikearaya:
Closes #582


Mikearaya:
don't mind the access control on deleteAccount API but  Should information about the user in order be masked such as `contact` and `billingAddress`?

pozylon:
@Mikearaya rebase

pozylon:
@Mikearaya rebase

pozylon:
I've rebased this again so it's compatible for a merge after localization changes

packages/api/src/schema/mutation.ts
```diff
@@ -884,6 +884,7 @@ export default [
       Remove user W3C push subscription object
       """
       removePushSubscription(p256dh: String!): User!
+      deleteAccount(userId: ID): Boolean!
```
pozylon, line 887: I don't like the naming "account" because we haven't used account across the whole system yet for API's. what about: 
- deleteUser ?

pozylon (changes requested):
- Remove enableRightToBeForgotten setting, due to international policies that should be the default and we want to be privacy friendly too. Every user should be able to delete him/herself

Then a general thing:
Order Data should stay immutable. Because of some laws it's usually required to keep users but delete user data. Thus we shouldn't delete orders and such but rather anonymize. What if we just delete all the profile data of the user:
- username
- services/password 
- e-mail from the user 
- last contact, heartbeat, IP's, last billing address
- avatar

Also, keep the reviews. Frontends then can decide to show the review of the user under some name like "Unknown User" or "Deleted User"

pozylon:
@Mikearaya we still need this feature but the branch has forked off far away from master now, i tried to rebase/merge but i think the easiest way is to generate a diff file based on the last ancestor (b827dbee) and then manually add the changes to a new branch based on master 🤪

Mikearaya:
@pozylon agreed, i'll recreate it based on v2.x

PR #589, opened 2024-04-05
Token validator

### LINKED ISSUE #588: [FEATURE REQUEST] Check Reset Token for Validity upfront
Currently we can't show if a reset link is expired until the moment when somebody wants to actually reset the password with the token. We should have another Mutation/Query to check if a token is valid so we can directly show an expired page when the token is not valid.

Mikearaya:
Added validation token verifier endpoint for both reset token and email verification token
- validateResetPasswordToken
- validateVerifyEmailToken

Closes #588


pozylon:
LGTM

PR #591, opened 2024-04-22
Fix duplicate assortment and product PUT based media upload

Mikearaya:
This PR will fix duplicate entries of assortment and product media when uploading using  PUT URL.

`registerFileUploadCallback` was being called multiple times when preparing and confirming PUT upload, which created a duplicate entry.


pozylon:
I'm trying to understand this case:

linkFile service is the only function that calls the registered hook. linkFile service is triggered in:
- Mutation.confirmMediaUpload
- in the webhook of Minio PUT event
- directly inside GridFS PUT method handler

Means to me:
- When using GridFS you never need to manually call confirmMediaUpload
- When using any S3/Minio unconfigured you need to call confirmMediaUpload manually but if you create a webhook pointed to unchained engine you don't need to do confirmMediaUpload.

Instead of trying to fix this inside the registered hooks, isn't it better to actually check if the media upload is confirmed before calling the hook?

linkFile does this too:
  await files.update(file._id, { size, type: type || file.type, expires: null });

so expires: null => ticket confirmed

pozylon:
I guess it's better to check in confirmMediaUpload if the media upload ticket is still valid (exists and has an expires date less than now). if the ticket does exist but has null expires -> throw an error saying the upload is already confirmed.
if the ticket does not exist or is expired -> throw an error saying the upload ticket has expired

That should solve the double calling, in the frontend, we just try to confirm after upload and do not show any error if we get the "already confirmed" error.

Another idea would be to not throw but just return the media when it's already confirmed instead of calling linkFile

packages/core-files/src/service/linkFileService.ts
```diff
@@ -7,7 +7,13 @@ export const linkFileService: LinkFileService = async ({ fileId, size, type }, u
...
-  await FileDirector.getFileUploadCallback(file.path)(file, unchainedAPI);
+
+  if (file.expires) {
+    await files.update(file._id, { size, type: type || file.type, expires: null });
+    const callback = FileDirector.getFileUploadCallback(file.path);
+    if (callback) {
+      await callback(file, unchainedAPI);
+    }
+  }
   return files.findFile({ fileId });
```
pozylon, line 18: findFile only necessary inside the if, else you can return the file from line 8 @Mikearaya

pozylon (changes requested):

packages/core-files/src/service/linkFileService.ts
pozylon: https://methodpoet.com/return-early-or-if-statement/

PR #593, opened 2024-05-01
Add target on externalLinks field 

Mikearaya:
… handled by admin-ui

### DIFF
--- a/packages/api/src/schema/types/shop.ts
+++ b/packages/api/src/schema/types/shop.ts
@@ -2,7 +2,7 @@ export default [
   /* GraphQL */ `
     enum ExternalLinkTarget {
       """
-      Open on new tab Enrollment
+      Open on new tab
       """
       BLANK
       """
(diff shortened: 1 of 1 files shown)


packages/api/src/schema/types/shop.ts
```diff
@@ -1,8 +1,19 @@
 export default [
   /* GraphQL */ `
+    enum ExternalLinkTarget {
+      """
+      Open on new tab Enrollment
```
pozylon, line 5: Enrollment?

pozylon (changes requested):

pozylon:
@Mikearaya please remove the "Enrollment", then merge and tell me to prepare an unchained version

packages/api/src/schema/types/shop.ts
Mikearaya: oopse

PR #596, opened 2024-05-15
Update createFilter api

Mikearaya:


packages/api/src/resolvers/mutations/filters/createFilter.ts
```diff
@@ -1,20 +1,23 @@
 import { log } from '@unchainedshop/logger';
 import { Root, Context } from '@unchainedshop/types/api.js';
-import { Filter } from '@unchainedshop/types/filters.js';
+import { Filter, FilterInputText } from '@unchainedshop/types/filters.js';
 
 export default async function createFilter(
   root: Root,
-  { filter }: { filter: Filter & { title: string } },
+  { filter, texts }: { filter: Filter; texts: Array<FilterInputText> },
```
pozylon, line 7: Naming is not consistent, here it's Array<FilterInputTexts> and for Assortments it's AssortmentText[].

Stick to one way, for ex. Array<AssortmentInputTexts> or FilterText[]

packages/api/src/resolvers/mutations/filters/createFilter.ts
```diff
@@ -1,20 +1,23 @@
...
+  const newFilter = await modules.filters.create(
     {
       ...filter,
+      title: '',
       locale: localeContext.language,
     },
     context,
   );
+  await modules.filters.texts.updateTexts({ filterId: newFilter._id }, texts);
+  return modules.filters.findFilter({ filterId: newFilter._id });
```
pozylon, line 22: You can savely return newFilter here (no need for extra request) because the texts are stored in another collection.

packages/api/src/resolvers/mutations/products/createProduct.ts
```diff
@@ -1,16 +1,15 @@
...
-  { modules, userId, localeContext }: Context,
+  { product }: { product: Product & { texts: ProductText[] } },
+  { modules, userId }: Context,
 ) {
   log('mutation createProduct', { userId });
 
   return modules.products.create({
     ...product,
-    locale: localeContext.language,
   });
```
pozylon, line 14: texts where updated?

pozylon (changes requested):
missing:
- product media text
- assortment media text
- product variation text
-

packages/api/src/resolvers/mutations/products/createProduct.ts
Mikearaya: yes

packages/api/src/resolvers/mutations/filters/createFilter.ts
```diff
@@ -1,20 +1,23 @@
 import { log } from '@unchainedshop/logger';
 import { Root, Context } from '@unchainedshop/types/api.js';
-import { Filter } from '@unchainedshop/types/filters.js';
+import { Filter, FilterInputText } from '@unchainedshop/types/filters.js';
 
 export default async function createFilter(
   root: Root,
-  { filter }: { filter: Filter & { title: string } },
+  { filter, texts }: { filter: Filter; texts: FilterInputText[] },
```
pozylon, line 7: texts should be optional types

packages/api/src/resolvers/mutations/filters/createFilter.ts
```diff
@@ -1,20 +1,23 @@
...
-  return modules.filters.create(
+  const newFilter = await modules.filters.create(
     {
       ...filter,
+      title: '',
       locale: localeContext.language,
     },
     context,
   );
+  await modules.filters.texts.updateTexts({ filterId: newFilter._id }, texts);
```
pozylon, line 21: Move this to mutation level and only do when texts supplied

packages/core-assortments/src/module/configureAssortmentsModule.ts
```diff
@@ -358,8 +357,8 @@ export const configureAssortmentsModule = async ({
         ...rest,
       });
 
-      if (locale) {
-        await assortmentTexts.updateTexts(assortmentId, [{ locale, title }]);
+      if (texts) {
+        await assortmentTexts.updateTexts(assortmentId, texts);
```
pozylon, line 361: Move this to mutation level and only do when texts supplied

packages/core-products/src/module/configureProductsModule.ts
```diff
@@ -341,8 +341,8 @@ export const configureProductsModule = async ({
         ...productData,
       });
 
-      if (locale) {
-        await productTexts.updateTexts(productId, [{ locale, title }]);
+      if (texts) {
+        await productTexts.updateTexts(productId, texts);
```
pozylon, line 345: Move this to mutation level and only do when texts supplied

packages/types/products.ts
```diff
@@ -296,7 +296,7 @@ export type ProductsModule = {
 
   // Mutations
   create: (
-    doc: Product & { title: string; locale: string },
+    doc: Product & { texts: ProductText[] },
```
pozylon, line 299: texts should be optional because the graphql field texts is optional too

pozylon (changes requested):
When looking at the code all those create module functions for assortments, products, filters actually should be decoupled from the texts now because what is happening currently is:

- Mutation takes optional texts
- Mutation forwards texts to create function
- create function checks for optional texts, then calls sub-module create function for texts

It should propably be:
- Mutation uses create function
- Mutation checks for optional texts supplied and uses texts create function

pozylon (changes requested):
Still not consistent,

1. all the mutations should take the texts as an addition parameter, not part of the input type of the product
2. it should always be of type [SomethingTextInput!] so i should be able to skip sending the parameter. Like it is now ]! it's required to send an empty array which sucks

pozylon:
Then what i saw additionally:
1. your linting config in vscode is not correct, code doesn't get linted and if you run npm run lint it shows you that

pozylon:
@Mikearaya not sure if this is ready yet? wasn't there some stuff missing like media texts?

pozylon:
Yes i see that, i checked and it was not because of media but for other things, API-wise it's not completely consistent yet, now it's:

      createProductVariation(productId: ID!, variation: CreateProductVariationInput!): ProductVariation!
      createProductVariationOption(
        productVariationId: ID!
        option: CreateProductVariationOptionInput!
      ): ProductVariation!

input CreateProductVariationOptionInput {
      value: String!
      title: String!
    }

input CreateProductVariationInput {
      key: String!
      type: ProductVariationType!
      title: String!
    }

so create product variation for ex. still has the old legacy way.

and create filter option has a title still:

input CreateFilterOptionInput {
      value: String!
      title: String!
    }


- for addProductMedia and addAssortmentMedia: can you add a @deprecated to the graphql schema and then don't change it?

PR #600, opened 2024-07-06
Admin status reporter

### LINKED ISSUE #578: Add a status API that returns monitoring-relevant information
- Amount of work items created, grouped by current status
- Amount of orders created, grouped by status
- Amount of events created

And show those numbers:
- For Today
- Since Start of Instance
- Since Ever

Mikearaya:
Closes #578


pozylon:
Cool:
```
type EventStatistics {
   event: EventType!
   count: Int!
}
```

Orders
```
type OrderStatistics {
   type: OrderStatus!
   count: Int!
}
```

pozylon:
Also lock down the statistics query with an action "viewStatistics" and allow it by default for admins only

Mikearaya:
adjusted the API based on the suggestions and added additional type parameters on event and work statistics API to filter by types.

packages/api/src/schema/query.ts
```diff
@@ -474,6 +474,19 @@ export default [
       Determines if a token is valid/active for email verification
       """
       validateVerifyEmailToken(token: String!): Boolean!
+
+      """
+      Returns aggregated report of all the events that occurred in the system
+      """
+      eventStatistics(types: [String!], from: Timestamp, to: Timestamp): [EventStatistics]!
```
pozylon, line 481: The case [Entity]! is almost never needed because it would include arrrays like this: [null, undefined]. Only in very rare cases we want this potential outcome.

99% of the time use [Entity!]! for array like return types in graphql

packages/api/src/schema/types/order/order.ts
```diff
@@ -81,5 +81,13 @@ export default [
       updated: DateTime
       user: User
     }
+
+    type OrderStatistics {
```
pozylon, line 85: Ask chatgpt to make these type names consistent, it's like a mix of past, present, verbs and nouns currently with no alignment.

Maybe use orderedCount instead of checkoutCount, confirmedCount and rejectedCount or:

- newCount
- checkoutCount
- rejectCount
- confirmCount
- fullfillCount

or:

- creationCount
- orderCount
- rejectionCount
- ..

you get the idea i guess

packages/api/src/schema/types/shop.ts
```diff
@@ -25,6 +25,12 @@ export default [
       customProperties: [AdminUiConfigCustomEntityInterface!]!
     }
 
+    type ActivityReport {
```
pozylon, line 28: delete?

packages/core-events/src/module/configureEventsModule.ts
```diff
@@ -81,5 +83,57 @@ export const configureEventsModule = async ({
       const count = await Events.countDocuments(buildFindSelector(query));
       return count;
     },
+    getReport: async ({ from, to, types } = { from: null, to: null, types: null }) => {
+      const pipeline = [];
+      const matchConditions = [];
+      if (from || to) {
```
pozylon, line 89: cool you're using pipelines! it's a little bit messy though, maybe some comments would help.

packages/core-orders/src/module/configureOrdersModule-queries.ts
```diff
@@ -22,6 +28,24 @@ export const buildFindSelector = ({ includeCarts, status, userId, queryString }:
   return selector;
 };
 
+const normalizeOrderAggregateResult = (data = []): OrderReport => {
+  const statusToFieldMap = {
```
pozylon, line 32: No, you should not map the status to count:

An order that has been created yesterday and has been fullfilled today should lead to createdCount +1 when the report does not cover today and fullfillfedCount +1 when the report includes today.

pozylon: Use the dates instead, there is:

  created: Date;
  confirmed?: Date;
  rejected?: Date;
  fullfilled?: Date;

pozylon (changes requested):
The aggregations are still not quite fine, check out my prior feedback, do not use current status, use the date fields to do statistics in a certain time range, we do not want a "balance sheet" of date X, we need a "p&l"-style report.

> List of status transitions based on the **timestamps** so we can do funneling

Mikearaya:
> a report covering today and yesterday would show:
> - created: 3
> - confirmed: 1
> - rejected: 1
> - fullfilled: 1
> that allows painting of burn-down and funneling charts.

@pozylon yes only data we will not be able to depict is order creation/PENDING and CART
because they don't have a timestamp. That is why i used status.

Maybe we can add a new timestamp **checkedout** in order? that will allow use to track the entire process from cart creation to fulfilled/rejected

pozylon:
There is a ordered timestamp that is set when a cart transitions from cart to pending (= checkout)

Mikearaya:
@pozylon Work report is still using status which is not consistent, should we follow the same patter as orders and have a timestamp for each status switch?

pozylon:
> @pozylon Work report is still using status which is not consistent, should we follow the same patter as orders and have a timestamp for each status switch?

I'm not sure if i understand correctly, in my opinion work report have enough timestamps to do it consistently?

There is already timestamps for creation, started and finished, right?

pozylon:
@Mikearaya I have harmonized the field names for consistency. But while testing i found out that the workStatistics still work the old way with the status mapping instead of going for the timestamps:

For example when I fire this query here:
```
 workStatistics {
    type
    newCount
    startedCount
    successCount
    errorCount
  }
```

the engine returns for the coinbase thing as an example:
```
{
        "type": "UPDATE_COINBASE_RATES",
        "newCount": 1,
        "startedCount": 0,
        "successCount": 852,
        "errorCount": 20
      }
```

But it should actually be:
```
{
        "type": "UPDATE_COINBASE_RATES",
        "newCount": 873,
        "startedCount": 872,
        "successCount": 852,
        "errorCount": 20
      }
```

Mikearaya:
please test now

pozylon:
LGTM

PR #603, opened 2024-08-19
Right to be forgotten

Mikearaya:


Mikearaya:
Use this branch to test https://github.com/xecutors/unchained-adminui/tree/user-deletion

pozylon:
@Mikearaya please check the requested changes on the old PR: https://github.com/unchainedshop/unchained/pull/581

We should not alter the orders after checkout, also reviews should stay by default and when the user has been marked deleted, a frontend can show the review as authored by a deleted user. I'd add a second mutation to delete all product reviews of a user.

So we have options:

1. remove my own or any (when admin) user but keep reviews
2. remove my own or any (when admin) user and also remove reviews
3. remove any (when admin) user's reviews later, no matter if the user is deleted or not

packages/core-users/src/module/configureUsersModule.ts
```diff
@@ -817,5 +852,14 @@ export const configureUsersModule = async ({
         {},
       );
     },
+    deleteUser: async ({ userId }, context) => {
+      const { modules } = context;
+      const { _id, ...user } = await findUserById(userId);
+      delete user?.services;
+      const maskedUserData = maskUserPropertyValues({ ...user, meta: null });
```
pozylon, line 859: This would lead to fields like updated and created having illegal values...

and generally speaking: we can't just set every property value to a sha hash because we can not know how that will effect the system that uses user extensions with custom code 😵‍💫

what if a hash of a username is beeing used as a username? this needs to be far more explicit.


this is an example of a user deletion extension from one customer
```
const moveUserToNirvana: VeloplusAuthAPI["moveUserToNirvana"] = async (
      userIdToDelete,
      loggedInUserId,
    ) => {
      await Users.updateOne(
        { _id: userIdToDelete },
        {
          $set: {
            username: `deleted-${new Date().getTime()}`,
            deleted: new Date(),
            deletedBy: loggedInUserId,
            emails: [],
            lastBillingAddress: {},
            services: {},
            roles: null,
            profile: null,
          },
        },
      );
    };
```


it should be similar to that: every default user property should be examined in detail and we should remove/unset/nullify all properties with PERSONAL DATA and only invent hashes for fields that are both required and have personal data (username beeing the only field). Properties that are purely metadata/informational and have no personal data value (for ex. an update date) can be left as is.

pozylon (changes requested):

pozylon:
Okay, i have many requests to make this perfect:

Keep the same naming of deletion mutation, thus the mutation should be called removeUser. That mutation already exists and it should be extended then to remove the reviews when the boolean flag has been set:

removeUser(removeUserReviews: Boolean = false)

In general, removeUser should always (in addition to removing/masking personal data):
- Remove carts of that user (and the cart's order positions, order delivery, order payment etc.)
- Remove login sessions
- Remove bookmarks of that user
- Remove draft/open quotations
- Remove draft/open enrollmens
- Remove payment credentials

You don't need to purge events or work queue because that data is transient and get's removed after a while automatically.

Also make sure to emit an event when the user has been deleted, that event should contain the old data of the user before masking.

Removing a user should not be allowed if there is any active subscription (enrollment), pending orders, quotations in status requested/processing, in that case you need to throw errors explaining the reason why they can't delete the user yet.

Mikearaya:
@pozylon we had this discussion here https://github.com/unchainedshop/unchained/pull/603#issuecomment-2301385467

pozylon:
@Mikearaya only those related entities belonging to an order in status open/null (= carts)

pozylon:
Currently "orders.delete" looks like this and is currently only triggered through removeOrder mutation which checks first if it's a cart or not.:

delete: async (orderId) => {
      const deletedCount = await mutations.delete(orderId);
      await emit('ORDER_REMOVE', { orderId });
      return deletedCount;
    },

which is bad, because it leaves order_* documents in the db with orderId's that don't exist anymore. So we should extend that function and call it within user deletion for all carts of that user.

pozylon:
> @pozylon ok there is no problem with deleting order payment and delivery but why delete order positions and orders when the user data is masked, that is why we masked the user instead if removing it completely right? we might need it for amalitics ir some part if the system might depend on it

pozylon:
ups, didn't want to close. i just had one more remark coming to my mind about this and our discussion on basecamp:

If a user to be deleted (no matter if guest or not) mets a range of conditions after all the mutations done (like masking, removing payment credentials, carts etc.), we can safely delete the document from the collection to get rid of the lonely user document.

To be a valid user candidate for complete removal, the user needs to match all of the following conditions:
- No reviews linked to the user left
- No orders (carts and checked out orders) linked to the user left
- No enrollments linked to the user left
- No quotations linked to the user left
- No tokens linked to the user left

pozylon:
@Mikearaya Went missing:

- Removing order positions of carts that get removed (i did that already in my last commit on this branch)
- Removing payment credentials of user -> We will fix this in a follow up ticket after merging with my BFR PR
- Removing WebAuthn stuff of user -> We will fix this in a follow up ticket after merging with my BFR PR
- If you delete a user, those stay logged in -> We will fix this in a follow up ticket after merging with my BFR PR

PR #605, opened 2024-09-03
Add status filter option to orders

Mikearaya:


pozylon (approved):
In which situation can the status be a string only when graphql input field is array? if it can't we should only support a status array, other than that this is a very useful extension.

Please base it on the v2.x branch though when you need it for publicare, v3 will not land at publicare this year.

Mikearaya:
@pozylon graphql only accepts array and since its just introduced we can be sure there will be no regression. But tge module function was working with string status and if any custom util was using that there will be a regression. Thats why i made it work both ways.

but i guess making adjustment to previous util is ok too so i will remove that

PR #606, opened 2024-11-09
Add extended user filters

Mikearaya:


packages/api/src/schema/inputTypes.ts
```diff
@@ -277,5 +277,26 @@ export default [
...
+    }
+    enum UserCartFilter {
+      HAS_CART
+      NO_CART
+    }
+
+    input UserFilterInput {
+      verificationStatus: UserVerificationFilter
+      loginWithinDays: DateFilterInput
+      cartItems: UserCartFilter
```
pozylon, line 298: I think the order and cart filters are somewhat arbitrary because naturally what you'd do is use Query.orders and return the users from there... so actually there should be a userId filter on the Orders List and not the opposite. So please do this for the moment DIRECTLY on the query params without the abstraction UserFilterInput:

lastLogin: DateFilterInput
emailVerified: Boolean

It's nice that you have found the DateFilterInput. We have used that already for Query.workQueue. Actually we should use that too for the reports like eventStatistics, orderStatistics, workStatistics)

pozylon: Which backend logic do you refer? I can't see additional queries in storefront for this case, it's just an order status dropdown that you pass to the order list here:
Bildschirmfoto 2024-11-13 um 11 23 30

and if you enter a userId top-right into "Search", we can see those users in the "User" column and jump from there? i think that should be enough.

pozylon:
orderCount doesn't seem to show up

packages/api/src/schema/inputTypes.ts
pozylon: You're right, that simplified solution would fall short about what the user wants to do in those cases. Nevertheless the changes will help him to go about 60-70% of the way.

On the other hand, i don't know if this kind of report belongs to the user list either, it's more like a special report. Somewhere we have to draw the line between "user management" and "reporting" where as user management should be the main focus of the admin ui.

pozylon: For example, we have many different requests for reporting-kind of requirements:

- Where can I see a list of orders that have a total order value of CHF X or more?
- What products have been added and removed between 01.01.2024 and 01.03.2024?
- What is the most used payment option?
- Can you send me a list of users that don't have a phone number set?
- Which products don't have a picture?
- Which categories are empty?
- What products miss a meta.gtin?
- ...

etc. etc. 

What those demands have in common is:
- It's usually some kind of aggregation between one or more collections with a desired grouped or limited outcome
- Many times based on custom fields
- Highly specific to customer individual e-commerce case


This is actually best done with a tool like:
- https://grafana.com/grafana/plugins/grafana-mongodb-datasource/
- https://redash.io
- https://www.metabase.com

Mikearaya:
Also Changed orderStatistcs, eventStatistics & workStatistics to use DateFilterInput

PR #608, opened 2024-11-26
Restrict file access

### LINKED ISSUE #592: Restricted file downloads
Currently unchained does not restrict access to a file, so once you get the URL of a media uploaded, that stays permanent and is only restricted by webserver rules.

I suggest extending this:
- When creating media upload tickets, allow to set it as a "private" file, signaling the file plugin to generate temporary download url's and not persistent ones
- Add "action.downloadFile" and wire it with the Media.url field resolver. By default it just checks for the file to be public and returns true, else it checks if the file has been uploaded by logged in user.
- Extend Media.url field resolver to generate a signed url through the file plugin when a file is "private"
- When accessing the url of a "Media" and the file is "private", generate a signature guarded url that expires based on fileSettings.privateFileSharingMaxAge o.e.
- Extend gridfs webhook to check if the file is public or if the provided signature works and the shared url is not expired.

The end goal is that one can overwrite the permission "action.downloadFile" and make it dependent on context:  It can be extended for any use case, for ex. completely lock it down and reject download access even for the uploader for form submission files for example.

Mikearaya:
Closes #592

### DIFF
--- a/packages/api/src/resolvers/mutations/assortments/prepareAssortmentMediaUpload.ts
+++ b/packages/api/src/resolvers/mutations/assortments/prepareAssortmentMediaUpload.ts
@@ -3,7 +3,11 @@ import { Context } from '../../../context.js';
 
 export default async function prepareAssortmentMediaUpload(
   root: never,
-  { mediaName, assortmentId }: { mediaName: string; assortmentId: string },
+  {
+    mediaName,
+    assortmentId,
+    asPrivate,
+  }: { mediaName: string; assortmentId: string; asPrivate?: boolean },
   context: Context,
 ) {
   const { services, userId } = context;
@@ -14,6 +18,7 @@ export default async function prepareAssortmentMediaUpload(
       directoryName: 'assortment-media',
       fileName: mediaName,
       meta: { assortmentId },
+      isPrivate: Boolean(asPrivate),
     },
     context,
   );
--- a/packages/api/src/resolvers/mutations/assortments/prepareAssortmentMediaUpload.ts
+++ b/packages/api/src/resolvers/mutations/assortments/prepareAssortmentMediaUpload.ts
@@ -3,11 +3,7 @@ import { Context } from '../../../context.js';
 
 export default async function prepareAssortmentMediaUpload(
   root: never,
-  {
-    mediaName,
-    assortmentId,
-    asPrivate,
-  }: { mediaName: string; assortmentId: string; asPrivate?: boolean },
+  { mediaName, assortmentId }: { mediaName: string; assortmentId: string; asPrivate?: boolean },
   context: Context,
 ) {
   const { services, userId } = context;
@@ -18,7 +14,6 @@ export default async function prepareAssortmentMediaUpload(
       directoryName: 'assortment-media',
       fileName: mediaName,
       meta: { assortmentId },
-      isPrivate: Boolean(asPrivate),
     },
     context,
   );
--- a/packages/api/src/resolvers/mutations/assortments/prepareAssortmentMediaUpload.ts
+++ b/packages/api/src/resolvers/mutations/assortments/prepareAssortmentMediaUpload.ts
@@ -3,7 +3,7 @@ import { Context } from '../../../context.js';
 
 export default async function prepareAssortmentMediaUpload(
   root: never,
-  { mediaName, assortmentId }: { mediaName: string; assortmentId: string; asPrivate?: boolean },
+  { mediaName, assortmentId }: { mediaName: string; assortmentId: string },
   context: Context,
 ) {
   const { services, userId } = context;
--- a/packages/api/src/resolvers/type/media-types.ts
+++ b/packages/api/src/resolvers/type/media-types.ts
@@ -10,7 +10,12 @@ export interface MediaHelperTypes {
 export const Media: MediaHelperTypes = {
   url: async (root, params, context) => {
     const { modules } = context;
-    await checkAction(context, actions.downloadFile, [root, params]);
-    return modules.files.getUrl(root, params);
+    try {
+      await checkAction(context, actions.downloadFile, [root, params]);
+      return modules.files.getUrl(root, params);
+    } catch (e) {
+      console.error(e);
+      return null;
+    }
   },
 };
--- a/packages/api/src/resolvers/type/media-types.ts
+++ b/packages/api/src/resolvers/type/media-types.ts
@@ -13,8 +13,8 @@ export const Media: MediaHelperTypes = {
       await checkAction(context, actions.downloadFile, [file, params]);
       const fileUploadAdapter = getFileAdapter();
       const mediaUrl = modules.files.getUrl(file, params);
-      if (file.isPrivate) {
-        return fileUploadAdapter.signUrl(mediaUrl, file._id);
+      if (file?.meta?.isPrivate) {
+        return fileUploadAdapter.createDownloadURL(mediaUrl, file._id);
       } else {
         return mediaUrl;
       }
--- a/packages/api/src/resolvers/type/media-types.ts
+++ b/packages/api/src/resolvers/type/media-types.ts
@@ -2,7 +2,6 @@ import { filesSettings, File as FileType, getFileAdapter } from '@unchainedshop/
 import { Context } from '../../context.js';
 import { checkAction } from '../../acl.js';
 import { actions } from '../../roles/index.js';
-
 export interface MediaHelperTypes {
   url: (language: FileType, params: Record<string, any>, context: Context) => Promise<string>;
 }
@@ -18,9 +17,7 @@ export const Media: MediaHelperTypes = {
         const expiryTimestamp = new Date(
           new Date().getTime() + (filesSettings?.privateFileSharingMaxAge || 0),
         ).getTime();
-
-        const mediaSignature = await fileUploadAdapter.signUrl(file?._id, expiryTimestamp);
-        return `${mediaUrl}?s=${mediaSignature}&e=${expiryTimestamp}`;
+        return fileUploadAdapter.signUrl(mediaUrl, file._id, expiryTimestamp);
       } else {
         return mediaUrl;
       }
--- a/packages/api/src/resolvers/type/media-types.ts
+++ b/packages/api/src/resolvers/type/media-types.ts
@@ -1,4 +1,4 @@
-import { filesSettings, File as FileType, getFileAdapter } from '@unchainedshop/core-files';
+import { File as FileType, getFileAdapter } from '@unchainedshop/core-files';
 import { Context } from '../../context.js';
 import { checkAction } from '../../acl.js';
 import { actions } from '../../roles/index.js';
@@ -14,10 +14,7 @@ export const Media: MediaHelperTypes = {
       const fileUploadAdapter = getFileAdapter();
       const mediaUrl = modules.files.getUrl(file, params);
       if (file.isPrivate) {
-        const expiryTimestamp = new Date(
-          new Date().getTime() + (filesSettings?.privateFileSharingMaxAge || 0),
-        ).getTime();
-        return fileUploadAdapter.signUrl(mediaUrl, file._id, expiryTimestamp);
+        return fileUploadAdapter.signUrl(mediaUrl, file._id);
       } else {
         return mediaUrl;
       }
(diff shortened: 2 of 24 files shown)


packages/api/src/resolvers/type/media-types.ts
```diff
@@ -1,12 +1,32 @@
...
 
 export const Media: MediaHelperTypes = {
-  url(root, params, { modules }) {
-    return modules.files.getUrl(root, params);
+  url: async (file, params, context) => {
+    const { modules } = context;
+    try {
+      await checkAction(context, actions.downloadFile, [file, params]);
+      const fileUploadAdapter = getFileAdapter();
+      const mediaUrl = modules.files.getUrl(file, params);
```
pozylon, line 16: getUrl should determine the URL and the File Adapter should be responsible to generate URL's that are accessible, else it won't work across different providers.  So that logic here for adding the signing query param should be moved to the adapter.

packages/api/src/roles/all.ts
```diff
@@ -30,6 +30,11 @@ export const all = (role, actions) => {
     return false;
   };
 
+  const isPrivateFile = (file, _, context) => {
+    if (!file?.isPrivate || file?.meta?.authorId === context?.userId) return true;
```
pozylon, line 34: 1. isPrivateFile is the wrong name for this function, it should be something like isFileAccessible because it also allows access to public files
2. We have only used the naming 'authorId' for reviews, we should user userId to stay consistent with the rest

packages/api/src/roles/all.ts
```diff
@@ -45,6 +50,7 @@ export const all = (role, actions) => {
   role.allow(actions.viewWarehousingInterfaces, () => false);
   role.allow(actions.viewOrders, () => false);
   role.allow(actions.sendEmail, () => false);
+  role.allow(actions.removeUser, () => false);
```
pozylon, line 53: wrong PR

packages/api/src/schema/mutation.ts
```diff
@@ -852,9 +852,17 @@ export default [
       """
       signPaymentProviderForCheckout(orderPaymentId: ID, transactionContext: JSON): String!
 
-      prepareProductMediaUpload(mediaName: String!, productId: ID!): MediaUploadTicket!
-      prepareAssortmentMediaUpload(mediaName: String!, assortmentId: ID!): MediaUploadTicket!
-      prepareUserAvatarUpload(mediaName: String!, userId: ID): MediaUploadTicket!
+      prepareProductMediaUpload(
+        mediaName: String!
+        productId: ID!
+        asPrivate: Boolean
```
pozylon, line 858: Looking at the other queries and mutations, we have never used 'as' or 'is' prefixing for the boolean parameters, so just go for private: Boolean here.

packages/api/src/schema/mutation.ts
```diff
@@ -852,9 +852,17 @@ export default [
...
+      ): MediaUploadTicket!
+      prepareAssortmentMediaUpload(
+        mediaName: String!
+        assortmentId: ID!
+        asPrivate: Boolean
```
pozylon, line 863: and here... and everywhere where you use asPrivate

packages/file-upload/src/director/FileAdapter.ts
```diff
@@ -2,13 +2,26 @@ import { Readable } from 'stream';
 import { log, LogLevel } from '@unchainedshop/logger';
 import { IBaseAdapter } from '@unchainedshop/utils';
 import { UploadedFile, UploadFileData } from '../types.js';
+import crypto from 'crypto';
+
+const signUrl = (fileUrl: string, expiry: number): string => {
```
pozylon, line 7: Move this to the gridfs adapter, only the adapter knows how to sign correctly

packages/plugins/src/files/gridfs/gridfs-adapter.ts
```diff
@@ -29,15 +29,15 @@ export const GridFSAdapter: IFileAdapter = {
 
   ...FileAdapter,
 
-  async createSignedURL(directoryName, fileName) {
+  async createSignedURL(directoryName, fileName, context, isPrivate) {
```
pozylon, line 32: createSignedURL is used for uploading stuff, the expiry date hashing is therefore already there (in buildHashedFilename) so i think there is no changed needed here?

packages/plugins/src/files/gridfs/gridfs-webhook.ts
```diff
@@ -77,8 +77,21 @@ export const gridfsHandler = async (
 
     if (req.method === 'GET') {
       const fileId = fileName;
+      const { s: signature, e: expiryTimestamp } = req.query;
 
       const file = await modules.gridfsFileUploads.getFileInfo(directoryName, fileId);
+      if (signature && expiryTimestamp) {
```
pozylon, line 83: Oh, security here: i can therefore just drop the query parameters and get infinite public access to that file? ;)

pozylon (changes requested):
Check my comments, nearly there. Other tasks that are missing:

1. This kind of stuff needs an integration test IMO where we check if the security holds up to it's promise (there is already tests for uploads)
2. Allthough we only want to support this feature for gridfs for the moment, if somebody uses minio as a file plugin and sets the private mode, we need to throw an error.
3. non-private uploads should only be allowed for admin's by default so it's the developer that needs to unlock public uploads by extending the permissions for a specific project
4. private downloads should only be allowed for non-guest users by default

pozylon:
something else what i was thinking about:

i mean it could make sense to let somebody upload an avatar image privately or use this feature in some client-project for forms and such.

But what is the case with assortment and product media? i think we should only support this feature in the avatar mutation and remove the graphql parameter from prepareAssortmentMediaUpload and prepareProductMediaUpload

pozylon (changes requested):
...

packages/core/src/services/createSignedURLService.ts
```diff
@@ -6,27 +6,29 @@ import {
...
     modules: { files },
   } = unchainedContext;
   const fileUploadAdapter = getFileAdapter();
-  const preparedFileData = await fileUploadAdapter.createSignedURL(
-    directoryName,
-    fileName,
-    unchainedContext,
-  );
-  const fileData = getFileFromFileData(preparedFileData, meta);
+  if (isPrivate && fileUploadAdapter.key !== 'shop.unchained.file-upload-plugin.gridfs')
```
pozylon, line 21: 🫣 what is the idea of adapters when the caller adds a conditional for a specific adapter?

packages/file-upload/src/director/FileAdapter.ts
```diff
@@ -28,6 +29,11 @@ export interface IFileAdapter<Context = unknown> extends IBaseAdapter {
   createDownloadStream: (file: UploadedFile, unchainedAPI: Context) => Promise<Readable>;
 }
 export const FileAdapter: Omit<IFileAdapter, 'key' | 'label' | 'version'> = {
+  signUrl() {
+    return new Promise<string | null>((resolve) => {
```
pozylon, line 33: If you throw here, only adapters that can sign url's (for private reasons for ex.) sign, and other's crash (like minio) for isPrivate files, right?

packages/api/src/resolvers/type/media-types.ts
pozylon: still not resolved @Mikearaya

pozylon (changes requested):

packages/api/src/schema/mutation.ts
pozylon: I see, yeah no if possible we should stick with a single word that describes the purpose if somehow possible, the as,is,set,get things is somehow alien to the current schema, what about **exclusive** ?

Maybe something like that?

packages/api/src/resolvers/mutations/assortments/prepareAssortmentMediaUpload.ts
```diff
@@ -3,7 +3,7 @@ import { Context } from '../../../context.js';
 
 export default async function prepareAssortmentMediaUpload(
   root: never,
-  { mediaName, assortmentId }: { mediaName: string; assortmentId: string },
+  { mediaName, assortmentId }: { mediaName: string; assortmentId: string; asPrivate?: boolean },
```
pozylon, line 6: asPrivate

packages/api/src/resolvers/type/media-types.ts
```diff
@@ -1,12 +1,26 @@
...
```
pozylon, line 15: Url Transformation should come after getting the original URL and signing from the provider. getUrl should be able to transform the URL to CDN or Thumbor.

packages/api/src/resolvers/type/media-types.ts
```diff
@@ -1,12 +1,26 @@
...
+      if (file.isPrivate) {
+        return fileUploadAdapter.signUrl(mediaUrl, file._id);
```
pozylon, line 17: i think we should get rid of the isPrivate and let the provider decide the way this is handled based on a meta property.

So this code should be more or less:

const url = Adapter.buildDownloadUrl()
return fileSettings.transformUrl(url)

please move the getUrl handling to the setting default fn.

packages/core/src/services/createSignedURLService.ts
```diff
@@ -6,27 +6,26 @@ import {
...
     modules: { files },
   } = unchainedContext;
   const fileUploadAdapter = getFileAdapter();
-  const preparedFileData = await fileUploadAdapter.createSignedURL(
-    directoryName,
-    fileName,
-    unchainedContext,
-  );
-  const fileData = getFileFromFileData(preparedFileData, meta);
+  const preparedFileData = await fileUploadAdapter.createSignedURL(directoryName, fileName);
```
pozylon, line 21: Maybe there is no need to adjust this service at all, isPrivate and userId could be in meta, userId is already in meta for avatars for ex..

packages/file-upload/src/director/FileAdapter.ts
```diff
@@ -28,6 +29,13 @@ export interface IFileAdapter<Context = unknown> extends IBaseAdapter {
   createDownloadStream: (file: UploadedFile, unchainedAPI: Context) => Promise<Readable>;
 }
 export const FileAdapter: Omit<IFileAdapter, 'key' | 'label' | 'version'> = {
+  signUrl() {
```
pozylon, line 32: createDownloadURL is a better name and makes clear what it needs to do

packages/plugins/src/files/gridfs/gridfs-webhook.ts
```diff
@@ -77,8 +77,26 @@ export const gridfsHandler = async (
 
     if (req.method === 'GET') {
       const fileId = fileName;
-
+      const { s: signature, e: expiryTimestamp } = req.query;
       const file = await modules.gridfsFileUploads.getFileInfo(directoryName, fileId);
+      const fileDocument = await modules.files.findFile({ fileId });
+      if (fileDocument?.isPrivate) {
```
pozylon, line 83: fileDocument.meta?.isPrivate

packages/plugins/src/files/gridfs/gridfs-adapter.ts
```diff
@@ -28,7 +30,22 @@ export const GridFSAdapter: IFileAdapter = {
   version: '1.0.0',
 
   ...FileAdapter,
+  async signUrl(fileUrl: string, mediaId: string, expiry?: number) {
```
pozylon, line 33: I think this function should take the whole file object instead of just the URL.

so params should be file and expiry

pozylon (changes requested):
Please clean this whole thing up and create integration tests, this should be an easy task doable until lunch.

packages/plugins/src/files/gridfs/gridfs-adapter.ts
Mikearaya: @pozylon, it will not work because in media-type, getUrl is called with both the file and any query parameter supported in the file.url, such as version, for example.
and we can determine that parameter when calling it inside webhook

Mikearaya: @pozylon  it will not work because in media-type, getUrl is called with both the file and any query parameter supported in the file.url, such as version, for example.
That parameter can not be determined when calling it inside the webhook.

pozylon:
Merged into #609

PR #609, opened 2024-12-10
BFR: Big Falcon Refactoring

pozylon:


pozylon:
- Removes entanglement between core modules
- Removes entanglement between core modules and core context
- Remove obsolete typing
- Reduce Public API
- Properly use services concept

PR #611, opened 2024-12-13
Use net order pricing and migrate persisted calculations

pozylon:
## Summary

Order category balances now exclude tax, matching the shared pricing-sheet arithmetic. Startup converts every legacy persisted order calculation, including carts and completed or rejected orders, before workers start and the platform returns ready to serve requests. The existing plugin and API initialization order is retained.

- Preserve recorded rows, tax amounts, timestamps and metadata without repricing. Add tax offsets and explicit precision adjustments so historical gross/net balances, half-cent totals and discount breakdowns retain their values. Verify the conversion before each atomic update; retries and concurrent recalculations are safe.
- Keep gross item revenue consistent in MCP sales summaries, monthly reports and customer-spending statistics for both migrated and newly calculated orders. Retain fractional discount entries that round to zero.
- Run migrations even when workers are disabled. Invalid or contradictory historical calculations stop startup with the order ID; failed startup closes initialized plugins, API resources and the database connection.

## Rollout

Stop older application instances before upgrading and back up the database. Rolling back to a gross-format release requires restoring that backup. Custom order adapters must pass net amounts with separate tax amounts; direct `initCore` integrations must execute registered migrations before serving requests. See `MIGRATION.md` for details.

## Validation

- Full unit and integration suite: 1,814 passed, four skipped, zero failures.
- TypeScript package build, ESLint on changed TypeScript files and whitespace checks passed.
- The two review reproductions now preserve the historical CHF 1.21 total and consistent gross MCP revenue.
- HTTP GraphQL contracts cover exact gross/net category prices, discount entities, unit prices, every order status, historical rounding, empty calculations and taxable cart checkout through registered plugins.
- Deterministic Saferpay and PostFinance contracts verify outbound amounts/currencies, payment validation and PostFinance refunds for fresh, migrated and historical rounding cases. Gateway boundaries are mocked; no credentials are needed.
- Ran the same 14 GraphQL/payment contract tests against both the pre-net `origin/master` revision (`e420ba74d`) and this branch: 14 passed on each, no skips. All monetary expectations are identical; only test fixture storage/input amounts adapt via `PRICING_CONTRACT_FORMAT=gross` on the baseline. The baseline has isolated workspace dependencies and build outputs.
- Independent differential verification matched 120,000 historical calculations against the old pricing implementation, including fractional values, multiple taxes and discounts, and every public pricing scope.

The integration setup registers the `all` preset, but does not exercise every bundled plugin. Optional EU/UK/US tax presets, currency conversion, mixed tax rates and discount spillover into nonzero fees still lack a complete cart integration matrix. Gateway contract tests mock the external service; they do not validate live provider behavior.

Validation also exposed an intermittent test isolation race: an enrollment notification worker can fork email work after the next test resets the database, leaving an extra `WORK_ADDED` event and failing six event-list/count assertions. This was reproduced independently; isolated events tests and a subsequent full run passed. No pricing assertion failed, and this PR does not change the unrelated test setup.


pozylon:
Rebased onto current `master` and split: the migration-execution / startup-barrier changes (run migrations before workers, abort startup on failure, teardown de-duplication) moved to #783, which does not belong to the net/gross pricing topic. This PR is now net/gross only — pricing sheet, adapters, the `20260907120000` data migration, sales-report reads, and tests.

PR #612, opened 2024-12-13
Keycloak test

pozylon:


pozylon:
This is just an experiment connecting unchained with a self-hosted Keycloak server.

PR #624, opened 2025-01-06
Product variation assignment fixs

Mikearaya:


pozylon:
What bug does this fix?

pozylon:
I see, I think we should fix this the other way round: The idea for the vectors is that those are exhaustive.

With that I mean that when you have 3 variations with 3 values each, the total amount of valid combinations should be 9 and every combination is unique. It should actually not be possible to assign two products to the same proxy with the exact same vector values. And it should only be possible to add an assignment when all vectors are provided.

But the way the _id is generated seems error prone anyways to me, as it uses 
Object.values(vectors). vectors is an unsorted object so it could generate the same _id for

{ variant1: 1, variant3: 1, variant2: 2 } and
{ variant1: 1, variant2: 1, variant1: 2 }

-> We need to sort the vector object by their keys first or add the actual variant keys to the _id like
proxyId:variant1:1:variant3:1:variant2:2
proxyId:variant1:1:variant2:1:variant3:2

Mikearaya:
@pozylon what do you men by "It should actually not be possible to assign two products to the same proxy with the exact same vector values. And it should only be possible to add an assignment when all vectors are provided."

let's say we have we a t-shirt store and color and size vectors but different brands Nike/Adidas etc... so with the above statement we should also create a brand vector to set the register 
sm-red-nike
sm-red-adidas
lg-blue-nike
lg-blue-adidas

does it look right?

what if we create color and size vectors and then assign brand shirts to those vectors?

pozylon:
I think we need a refactor of this function here: https://github.com/unchainedshop/unchained/blob/2d1f6341e216da77ef318ea73f3c8c47d8a58707/packages/core-products/src/module/configureProductsModule.ts#L436

It should only $push when no product for that vector exists already in proxy.assignments

And here we should check if the vector is "complete" and has valid values based on the variations, else it should throw: https://github.com/unchainedshop/unchained/blob/2d1f6341e216da77ef318ea73f3c8c47d8a58707/packages/api/src/resolvers/mutations/products/addProductAssignment.ts#L8

That way if I call the mutation twice with the same values, it will not lead to a double assignment and if I call the mutation not providing a value for every variation, it goes error.

pozylon:
> @pozylon what do you men by "It should actually not be possible to assign two products to the same proxy with the exact same vector values. And it should only be possible to add an assignment when all vectors are provided."
> 
> let's say we have we a t-shirt store and color and size vectors but different brands Nike/Adidas etc... so with the above statement we should also create a brand vector to set the register sm-red-nike sm-red-adidas lg-blue-nike lg-blue-adidas
> 
> does it look right?
> 
> what if we create color and size vectors and then assign brand shirts to those vectors?

Well, different t-shirts of different brands are usually not in the same proxy because it's not just a variant but another article in general.

But yes, it looks right, if we have variations:

size: l, m, g
color: blue, red
cut: v-neck, round-neck

It means we can only have a product each for every combination:

- l-blue-v-neck
- l-blue-round-neck
- l-blue-v-neck
- l-blue-round-neck
- l-red-v-neck
- l-red-round-neck
- l-red-v-neck
- l-red-round-neck
- m-blue-v-neck
- m-blue-round-neck
- m-blue-v-neck
- m-blue-round-neck
- m-red-v-neck
- m-red-round-neck
- m-red-v-neck
- m-red-round-neck
- s-blue-v-neck
- s-blue-round-neck
- s-blue-v-neck
- s-blue-round-neck
- s-red-v-neck
- s-red-round-neck
- s-red-v-neck
- s-red-round-neck

pozylon:
I can't have 2 products with s-red-round-neck

pozylon:
What happens If i remove a variation?
What happens If I add another variation?

-> in those cases, the admin UI should warn the user that he needs to re-do the assignments and all assignments should be killed during those mutations


If i only add a new variation value -> all good, there is just un-assigned combinations in assignments which is fine
If i remove a variation value -> we should remove assignments with that vector value, optionally warn in the admin ui when there already exist assignments for that value.

pozylon:
@Mikearaya Can you please take on this topic? maybe write integration tests first so we can check against the behavior as described. this definitely needs tests.

Mikearaya:
@pozylon  Added tests that check if vector is valid and also is complete so if the variations are

```
const variationMatrix = [
  { color: '#ff0000', size: 'sm' },
  { color: '#ff0000', size: 'lg' },
  { color: '#000000', size: 'sm' },
  { color: '#000000', size: 'lg' }
];
```
this will be wrong and throw error because  its incomplete

```
  { color: '#000000' }
``

and this will pass


```
  { color: '#000000', size: 'lg' }
``

please test

pozylon:
I rebased this on the most recent master

PR #650, opened 2025-07-28
EXtend order statistics

Mikearaya:


packages/core-orders/src/module/configureOrdersModule-queries.ts
```diff
@@ -202,6 +209,51 @@ export const configureOrdersModuleQueries = ({ Orders }: { Orders: mongodb.Colle
...
+        { $unwind: '$calculation' },
+        {
+          $group: {
+            _id: {
+              date: {
+                $dateToString: { format: '%Y-%m-%d', date: `$${field}` },
+              },
+              currency: '$currencyCode',
+            },
+            totalAmount: { $sum: '$calculation.amount' },
```
pozylon, line 240: That won't work, we can't use aggregate here, we should return Orders with the date field, calculation, currencyCode. That way you can accept further input to the total field and resolve accordingly:

```
total(order: OrderType, params: { category: string; useNetPrice: boolean }): Price {
    const pricing = OrderPricingSheet({
      calculation: order.calculation,
      currencyCode: order.currencyCode,
    });

    if (pricing.isValid()) {
      return pricing.total(params);
    }
    return null;
  },
```

pozylon (changes requested):

PR #657, opened 2025-09-15
autocomplete tags

### LINKED ISSUE #441: [FEATURE REQUEST] Tag Autocompletion
Add a query to get a list of all tags over all products

Mikearaya:
Closes #441


pozylon (changes requested):
Add this to your solution please:

1. Allow to set the tags via adminUiConfig property in startPlatform. Env variables should not be the only option to configure this.
2. If you frame it "product tags" also make sure to name it more explicit in the ShopInfo: productTags.
3. Also support configuring the assortment tags as tagging assortments is similar important to products because we allow filtering by it in Query.assortments like in Query.products

packages/core-assortments/src/module/configureAssortmentsModule.ts
```diff
@@ -514,6 +514,10 @@ export const configureAssortmentsModule = async ({
     links: assortmentLinks,
     products: assortmentProducts,
     texts: assortmentTexts,
+    existingTags: async (): Promise<string[]> => {
+      const tags = await Assortments.distinct('tags', { tags: { $exists: true } });
```
pozylon, line 518: Only consider Assortments that are not deleted

packages/core-products/src/module/configureProductsModule.ts
```diff
@@ -600,6 +600,10 @@ export const configureProductsModule = async ({
     },
 
     texts: productTexts,
+    existingTags: async (): Promise<string[]> => {
+      const tags = await Products.distinct('tags', { tags: { $exists: true } });
```
pozylon, line 604: Only consider Products that are not deleted

pozylon (changes requested):
Almost there, only some fine grained suggestions left

PR #667, opened 2025-10-19
Fix add product item filter

Mikearaya:
Currently, when adding an existing product in the cart multiple times, it will add a new OrderItem entry instead of increasing the quantity, causing the issue mentioned here. https://github.com/xecutors/age-verified-shop/pull/3 This PR fixes it


pozylon:
Nice, tests pass

PR #670, opened 2025-10-21
Fix leveled catalog price calculations

Mikearaya:
it was not picking the right price for the current cart quantity of a product; it was defaulting to the first to the first price  because of !p.maxQuantity

let say 
```
const pricing = [
  { maxQuantity: 0, price: 300 },
  { maxQuantity: 5, price: 250 },
  { maxQuantity: 10, price: 200 },
];
```
Now the price calculation will be as follows

quantity   price
1    =>          300
4        =>        300
5          =>       250
9           =>     250
10       =>      200
15      =>      200

but before it was always 300 no matter the quantity

### DIFF
--- a/packages/core-products/src/module/configureProductPrices.ts
+++ b/packages/core-products/src/module/configureProductPrices.ts
@@ -69,7 +69,7 @@ export const configureProductPricesModule = ({
     });
 
     const foundPrice = pricing.reduce(
-      (current, tier) => (tier.maxQuantity != null && quantity >= tier.maxQuantity ? tier : current),
+      (current, tier) => (tier.maxQuantity && quantity >= tier.maxQuantity ? tier : current),
       pricing?.[0],
     );
     if (!foundPrice) return null;


Mikearaya:
@pozylon 
even in that case i think the implementation was wrong



```
const pricing = [
  { maxQuantity: 5, price: 250 },
  { maxQuantity: 10, price: 200 },
  { maxQuantity: 0, price: 300 },
];
const foundPrice = pricing.find((level) => !level.maxQuantity || level.maxQuantity >= quantity);
```

in this case if quantity is 11 then it will default to 0  => 300 when it should have matched 10 => 200

pozylon:
Adjusting it to minQuantity is unchained v5 because it's very breaking, i'll push a change with the fix for the explicit null case and fixed tests and close this PR.

pozylon:
@Mikearaya can you extend the tests so we test the catalogPrice calculation for these cases:

- order item quantity is 1 and matches the first explicit maxQuantity (pricing record 1 price)
- order item quantity matches one of the maxQuantity's (pricing record 2 price)
- order item quantity is higher than any of the maxQuantity's (should use price of null/0/undefined maxQuantity / pricing record 3 price)

pozylon:
Fix is pushed

pozylon:
Yeah, let's have a small issue for v5

PR #675, opened 2025-11-10
DeliveryProvider pickupLocations resolver

### LINKED ISSUE #656: Up-front display of Pick up locations
In order to get a list of pickup locations you have to set the delivery provider on the cart first and then query the locations on that OrderDelivery.

We should find a way to let the user query for pickup locations baed on the PROVIDER and not on the order.
-> DeliveryProvider.pickUpLocations instead of OrderDelivery.pickUpLocations

Mikearaya:
Closes #656 

OrderDelivery.pickupLocations still available, for convenience and backward compatibility

### DIFF
--- a/packages/api/src/schema/types/order/pick-up-location.ts
+++ b/packages/api/src/schema/types/order/pick-up-location.ts
@@ -1,12 +1,5 @@
 export default [
   /* GraphQL */ `
-    type OrderPickUpLocation {
-      _id: ID!
-      order: Order!
-      name: String!
-      address: Address
-      geoPoint: GeoPosition
-    }
     type PickUpLocation {
       _id: ID!
       name: String!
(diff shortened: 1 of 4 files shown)


packages/api/src/schema/types/order/pick-up-location.ts
```diff
@@ -7,5 +7,11 @@ export default [
       address: Address
       geoPoint: GeoPosition
     }
+    type PickUpLocation {
```
pozylon: I think it's not needed to have an Order field on the OrderPickUpLocation so please remove the old type and re-use PickUpLocation in OrderDeliveryPickUp too. Aaand, i think pickUpLocations on OrderDeliveryPickUp should be marked deprecated in the graphql schema.

PR #676, opened 2025-11-11
Default pdf templates

Mikearaya:


examples/kitchensink/src/boot.ts
```diff
@@ -12,6 +12,11 @@ import '@unchainedshop/plugins/pricing/discount-half-price-manual.js';
 import '@unchainedshop/plugins/pricing/discount-100-off.js';
 import { registerProductDiscoverabilityFilter } from '@unchainedshop/core';
 
+
+export const modules = {
+  ...defaultModules,
```
pozylon: bloat

examples/ticketing/.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 .env
 lib
-.db
\ No newline at end of file
+.db
+gastro-zurich*.json
```
pozylon: client name exposed inn git -> rm

packages/ticketing/src/optional.d.ts
```diff
@@ -0,0 +1,20 @@
+// src/types/optional.d.ts
+declare module '@react-pdf/renderer' {
```
pozylon: That file is not needed, put those dependencies into dev dependences of ticketing and optional dependencies, that way you can access the types safely.

packages/ticketing/src/pdf-tickets/googleWalletPass.ts
```diff
@@ -0,0 +1,399 @@
...
+    const token = this.jwt.sign(claims, this.credentials.private_key, {
+      algorithm: 'RS256',
+    });
+
+    return `https://pay.google.com/gp/v/save/${token}`;
+  }
+}
+
+export default async (token, unchainedAPI: Context) => {
+  const GoogleWallet = new GoogleEventTicketWallet(process.env.GOOGLE_WALLET_ISSUER_ID as string);
```
pozylon: Import the deps at the top with top level await then test if deps are installed here, else throw here so that the server doesn't boot up when dev tries to use the google wallet renderer.

packages/ticketing/src/pdf-tickets/defaultOrderPDFRenderer.ts
```diff
@@ -0,0 +1,105 @@
+import { TicketingAPI } from '../types.js';
```
pozylon: we should call this file defaultTicketReceiptRenderer and change the params/fn names accordingly now we also support returning other media types.

Mikearaya:
I haven't tested the apple wallet pass, please do on your end if you can

pozylon (changes requested):
See my other comment, although the way you structured it comes together well, the actual "value" for the developer is not leveraged

pozylon:
@Mikearaya i have rebased this branch on master and did some code cleanups and added default initialization logic to the ticketing example with claude so make sure to hard reset before continuing your work. When done, we should be able to cleanup the Theater im Hof Repository substantially leveraging the factory functions to implement the simple ticket case.

Please make sure the tests pass in the ticketing example and use that to test ticketing stuff. Somehow we need to be able to easily test the processes there so it should seed a product that is a ticket and do some checkout and everything.

Mikearaya:
I think this is not needed anymore because it is part of master already

PR #678, opened 2025-11-17
Leveled catalog pricing Refactor

### LINKED ISSUE #671: Change leveled catalog price to use minQuantity logic
Continued from #670

Mikearaya:
Closes #671


Mikearaya:
@pozylon we might not be able to include this because of the breaking changes, but I've made all the changes we discussed in the other PR.
There is no maxQuantity usage anymore, and it's minQuantity-based, so the price specified for that quantity will only be applied when the order item is equal to or greater than the quantity.

pozylon:
Seems fine to me but you are right, we can't merge it until v5 and if we merge, we need to support the maxQuantity way in the bulk importer and marke it deprecated. That beeing said, the bulk importer should be able to convert a maxQuantity denoted pricing to a minQuantity denoted pricing before updating/inserting that commerce data.

Mikearaya:
ok, I will make that adjustment and keep this PR updated until it can be merged

PR #684, opened 2025-12-02
Show existing tags where tags are used

Mikearaya:
## Summary

- remove the standalone tags page and navigation entry
- show configured existing tags directly in product, assortment, and user tag inputs
- use the shared tag picker for bulk add and remove actions
- prefer matching existing tags and prevent case-only duplicates

## Verification

- ESLint
- TypeScript type check
- Cypress component tests: 10 passing
- Admin UI production build


Mikearaya:
its a continuation from archived repo, i dont know if we should also include this or not

PR #686, opened 2025-12-16
Bulk exporter

Mikearaya:


pozylon:
Seems to come together nicely! @fliptation where could we display the exports?

Currently it's triggered like this:

Bildschirmfoto 2025-12-18 um 21 37 00


And then it goes to:
Bildschirmfoto 2025-12-18 um 21 37 26

Propably we should find some place where we can show the exports that have been queued up with direct links the downloadable export files in a nice ui? notifications? dropdown? where could we put this?

Data wise it's more like:

18.12.2025 15:34 | Products, Product Variations | Generating Report...
17.12.2025 15:34 | Products, Product Texts | **Download Report**

pozylon:
Not sure if this is because of my rebase or refactoring of claude, but after i triggered an export it switches to the Recent Exports but doesn't show any entries. And it shows 0 (ui glitch)
Bildschirmfoto 2026-01-05 um 13 02 11

pozylon:
Bildschirmfoto 2026-01-05 um 13 02 43
bulk export shows up in the work queue though

pozylon (changes requested):
Looks good but i think it needs some testing because admin ui is not working correctly. please check my other comments

PR #691, opened 2026-02-26
Adding download to delivery provider list

networkinss:
I added "Download" to the list of delivery providers.


networkinss:
Seems you have trouble to make a decision. Are there open questions ?

PR #717, opened 2026-07-06
Migrate Admin UI plugins to native ESM and support local extension development

Mikearaya:
Admin UI plugins now load as native ES modules through a shared browser import map. Plugin builds emit `dist/index.js`; Express and Fastify serve the plugin bundles and packaged SDK, preserve route-specific HTML, and support CSP nonces. Existing plugins must be rebuilt and their `bundlePath` changed from `.global.js` to `.js`.

Client projects can develop extensions against the prebuilt Admin UI package. Both Kitchensink examples build their local Bookmark Manager before backend startup and watch its source alongside the backend; refreshing the browser loads the rebuilt extension. The examples and SDK documentation describe setup and the distinction from developing the Admin UI through Next.js.

The change also validates plugin exports and SDK versions, keeps optional Admin UI dependencies lazy, adds runtime and HTML-routing tests, adjusts bookmark search input spacing, and includes the workspace's dependency lockfile refresh.

Validation: 729 unit tests pass; both Bookmark Manager builds pass; plugin watch rebuilding verified. Full Admin UI production build, including SDK bundles and TypeScript declarations, passes after merging the latest master.

CodeQL identified potentially expensive route-normalization regexes. Trailing slash removal now uses a linear scan, with regression coverage for long slash-heavy paths and mount prefixes. Targeted ESLint checks pass.


Mikearaya:
Also added a sortOrder to nav items so that the plugin-injected side navigation items can be displayed anywhere in the side nav

Mikearaya:
@pozylon  I changed my mind on this.

Earlier, I was leaning toward the IIFE approach because it was simpler, and ESM kept running into issues I couldn't explain. But after digging deeper, those turned out to be a Next.js build/dev cache bug that affects the IIFE branch too. ESM was never the real issue.
What really changed my mind is the long-term maintenance. IIFE relies on manually keeping export lists in sync, which is easy to forget and only fails at runtime. ESM handles all of that automatically, and plugins can share the host's SDK instead of bundling their own copy.

**IIFE is simpler to get started with, but I think ESM is simpler to live with. The migration is already done, please review and merge :)**

PR #719, opened 2026-07-09
Admin UI user cart and session management

Mikearaya:
Add user cart tab and logout-all-sessions support to admin-ui

- Show the active cart of a user on the user detail page via the
  user.cart field, with items, totals, and an admin-only link to
  the order detail view
- Add a Sessions section to the account view: "Log out everywhere"
  for the own account, "Force logout" for admins on other users
- Extend the logoutAllSessions mutation with an optional userId so
  admins can invalidate another user's sessions; gated by the
  existing isMyself ACL check and covered by integration tests


pozylon:
Bildschirmfoto 2026-09-02 um 12 11 29
needs improvement and doesn't show the sessions ;)

pozylon:
ah, sorry if you can show the session list, if you can't because we don't have that feature (not sure), then it's fine and it's just the buttton that looks crap

pozylon:
on http://localhost:3000/users/?userId=d2414eb64a80b2c085750688&tab=account i got this:
Bildschirmfoto 2026-09-02 um 12 14 59

Mikearaya:
@pozylon you are right; there is no way to show active sessions of a user yet

PR #720, opened 2026-07-09
Mcp optimization

Mikearaya:


pozylon:
Superseded by #777 after reconciling this branch with the stateless SDK-v2 MCP architecture now on `master`.

The session/client/resource cache in this PR is no longer valid: most of its surrounding work has already landed through the SDK-v2 migration, while retaining live auth-bound clients would reintroduce security and lifecycle risks. A three-round benchmark (25 warm-ups + 200 samples per workload) measured current MCP setup at 6.901 ms p95 and a metadata-only prototype at 2.806 ms p95. The 4.095 ms saving missed the agreed 20 ms absolute threshold, so #777 intentionally adds no cache.

The one still-valid change—the corrected `users_management` description—has been retained in #777 with integration coverage. Closing this PR without deleting its branch so the original discussion and implementation remain available for history.

Issue #722, opened 2026-08-19
FiltersSettings cache contract cannot express pruning, so custom backends reintroduce #721


pozylon:
**Follow-up to #721.**

The default MongoDB product-id cache now prunes obsolete filter option rows, and deleting a
filter drops its whole cache. Both live in the Mongo backend, and the option-level half of it
cannot be expressed through the `FiltersSettings` contract — so a shop that swaps the cache
backend silently gets #721 back.

**Where the seam falls short**

```ts
setCachedProductIds: (
  filterId: string,
  productIds: string[],
  productIdsMap: Record<string, string[]>,
) => Promise<number>;
```

Nothing here says *"these are the only valid keys, drop everything else"*. The Mongo backend
works around it by reading the `Filters` collection directly and pruning against the filter's
current options. A Redis / HTTP / in-memory backend has no such access, so it can only ever
add keys and never retire them — exactly the shape of #721.

**Why the map itself is not the answer**

Pruning against `Object.keys(productIdsMap)` looks like the obvious fix, but that map is a
snapshot taken at the *start* of a full invalidation. `FilterDirector.buildProductIdMap` scans
the catalog once per option, sequentially, so on a large catalog it is in flight for minutes.
If an option is added during that window, the slow invalidation lands last carrying the older
option set, and pruning against it deletes the new option's cache row. A *current, valid*
option then resolves to nothing and disappears from the facet list — a louder failure than the
stale rows #721 is about. The Mongo backend avoids this by resolving the authoritative option
set as late as possible, at write time.

**Directions worth weighing**

1. Hand the backend the authoritative value set, resolved as late as possible, rather than
   letting it re-derive one.
2. Make replacement semantics explicit: `setCachedProductIds` means "this is the complete cache
   for this filter", and backends are required to drop keys not present.
3. Keep pruning in `FilterDirector` and give the settings a per-key delete primitive.

(1) and (2) both keep the write and the prune in one round trip, which matters because this
runs for every filter on boot and after every bulk import.

**Related rough edge**

`FiltersSettings` members each fall back to the Mongo default independently:

```ts
filtersSettings.setCachedProductIds = setCachedProductIds || defaultCache.setCachedProductIds;
filtersSettings.getCachedProductIds = getCachedProductIds || defaultCache.getCachedProductIds;
filtersSettings.purgeCachedProductIds = purgeCachedProductIds || defaultCache.purgeCachedProductIds;
```

[...]

Issue #721, opened 2026-08-19
Filter product id cache keeps rows of removed filter options, so obsolete option values keep returning frozen results


pozylon:
**Describe the bug**

`filter_productId_cache` rows are only ever upserted, never removed. When a filter option
disappears (renamed, removed, or the whole set of option values replaced), its cache row stays
behind — and because the lookup map is rebuilt from *all* rows of a filter, the obsolete value
keeps resolving. It answers with the product ids it had at the moment the option was removed,
frozen from then on.

The two halves of it:

- [`setCachedProductIds`](https://github.com/unchainedshop/unchained/blob/master/packages/core-filters/src/product-cache/mongodb.ts#L62-L79) writes the base row plus one row per entry of `productIdsMap`, via an upsert. Nothing deletes.
- [`getCachedProductIds`](https://github.com/unchainedshop/unchained/blob/master/packages/core-filters/src/product-cache/mongodb.ts#L35-L56) builds `productIdsMap` from every row it finds for that `filterId`, so leftovers become valid lookup keys again.

Re-invalidating does not help: [`FilterDirector.buildProductIdMap`](https://github.com/unchainedshop/unchained/blob/master/packages/core/src/directors/FilterDirector.ts#L167-L185) derives the map from `filter.options`
(or `true`/`false` for `SWITCH`), so a value that is no longer an option is simply absent from
the map — and absent means "not written", not "removed".

`grep -rn "FilterProductIdCache" packages/` shows the collection is only read from and upserted
into; there is no `deleteMany`/`deleteOne`/`$pull` against it anywhere in the repo. The same
applies when a filter is deleted entirely: [`modules.filters.delete`](https://github.com/unchainedshop/unchained/blob/master/packages/core-filters/src/module/configureFiltersModule.ts) removes the filter and its
texts, but leaves its cache rows.

What makes this hard to spot from the outside: a value that was *never* an option correctly
returns nothing, so the filter looks like it works and merely returns "too few" products.

**To Reproduce**

Self-contained script — only `@unchainedshop/core-filters`, `@unchainedshop/core`,
`mongodb-memory-server` and `mongodb`:

```js
import { MongoMemoryServer } from 'mongodb-memory-server';
import { MongoClient } from 'mongodb';
import { configureFiltersModule, filtersSettings } from '@unchainedshop/core-filters';
import { FilterDirector } from '@unchainedshop/core';

const mongod = await MongoMemoryServer.create();
[...]

PR #723, opened 2026-08-19
fix(core-filters): prune product id cache rows of removed filter options (port from v4.8.x)

pozylon:
Ports the #721 fix from `v4.8.x` (30841bb15) to `master`.

## What was wrong

`setCachedProductIds` only ever upserted, so the cache row of a filter option that got renamed
or removed stayed behind. The lookup map is rebuilt from *every* row of a filter, so the
obsolete value kept resolving to the product ids it had at the moment it was retired, frozen
from then on. Re-invalidating did not help: the map is derived from `filter.options`, so a
value that is no longer an option is simply absent — and absent meant "not written", not
"removed".

RANGE filters are hit harder than #721 describes: their parser resolves `start:end` against the
cached keys, so a retired bucket is picked up by an *ordinary* range query, not only by a query
for the obsolete value itself.

## Notes on the approach

The prune resolves the authoritative option set by re-reading the filter at write time rather
than trusting the `productIdsMap` it was handed. That map is a snapshot taken at the start of a
full invalidation, which scans the catalog once per option and can be in flight for minutes;
pruning against it deletes options added in the meantime, turning a stale row into a missing
one so that a *current* option silently resolves to nothing — a louder failure than the bug
being fixed. There is a regression test for exactly that case.

Deleting a filter drops its cache through the new `filtersSettings.purgeCachedProductIds`, so
the module keeps talking to the settings seam instead of reaching into the cache collection.
Option-level pruning still cannot be expressed through that contract, so custom cache backends
keep the old behaviour — tracked in #722.

## Conflict resolved during the port

`product-cache/mongodb.ts` only — `master` uses `memoizeWithTTL` where `v4.8.x` uses
`pMemoize` + `ExpiryMap`. Kept master's memoization, added the `filterOptionValues` import.

## Verification on master

- Integration: **1001 pass, 0 fail** (ran twice)
- Unit: **589 pass, 0 fail**
- Build + ESLint clean
- Reverting only the `packages/` changes makes 4 of the 5 new tests fail, confirming they guard
  the fix. The fifth is the concurrency guard, which passes on unfixed code by construction —
  without a prune nothing is ever deleted.

One test needed a fix during the port that `v4.8.x` did not surface: cached reads are memoized
in process with a 1 ms TTL outside production, and `memoizeWithTTL` expires on
`Date.now() > expiry`, so two reads inside the same millisecond both hit the memo. The test now
waits that out before asserting.

### DIFF
--- a/packages/core-filters/src/db/FiltersCollection.ts
+++ b/packages/core-filters/src/db/FiltersCollection.ts
@@ -18,6 +18,14 @@ export type Filter = {
   type: FilterType;
 } & TimestampFields;
 
+/*
+ * The set of values a filter can be queried by, and thus the keys of its product id cache:
+ * a SWITCH is not backed by options but by the two boolean states, every other type is
+ * defined by its options.
+ */
+export const filterOptionValues = (filter: Pick<Filter, 'type' | 'options'>): string[] =>
+  filter.type === FilterType.SWITCH ? ['true', 'false'] : filter.options || [];
+
 export type FilterText = {
   _id: string;
   filterId: string;
--- a/packages/core-filters/src/filters-settings.ts
+++ b/packages/core-filters/src/filters-settings.ts
@@ -8,6 +8,8 @@ export interface FiltersSettings {
     productIdsMap: Record<string, string[]>,
   ) => Promise<number>;
   getCachedProductIds: (filterId: string) => Promise<[string[], Record<string, string[]>] | null>;
+  /* Drop the whole cache of a filter, used when the filter itself goes away. */
+  purgeCachedProductIds: (filterId: string) => Promise<void>;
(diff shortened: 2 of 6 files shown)


pozylon:
Closing in favour of cherry-picking from `v4.8.x` directly.

This branch is also superseded: it carries the earlier shape where the MongoDB backend re-read `Filters` to work out which option values were still current. `v4.8.x` has since moved that responsibility to `FilterDirector.invalidateProductIdCache` (d0715dcd2), which discards results a concurrent option change has superseded and lets `setCachedProductIds` simply replace what it is given. Merging this as-is would put an approach on master that v4.8.x has already left behind.

The work lives on `v4.8.x` as 30841bb15 + 2b9d82807 + d0715dcd2.

PR #724, opened 2026-08-20
fix(core-worker): stop autoscheduled work from being queued with a null input (port from v4.8.x)

pozylon:
Ports 0308e31bf from `v4.8.x`.

## What was wrong

`ensureOneWork` passed `input` straight into `$set`. An undefined value is serialized as null
rather than skipped, so **every autoscheduled work item was stored with `input: null`**.

Adapters that destructure their input with a `= {}` default then threw, because a default
parameter only guards `undefined`:

```
Cannot destructure property 'guestUserMaxAgeInDays' of
'(intermediate value)(intermediate value)(intermediate value)' as it is null.
```

The correlation on a live instance is exact — the only autoscheduled adapters that destructure
are the only ones that ever failed:

| Type | signature | runs |
|---|---|---|
| `GC_GUESTS` | `({ guestUserMaxAgeInDays } = {})` | **0 ok / 11 failed** |
| `INVALIDATE_CARTS` | `({ maxAgeDays } = {})` | **0 ok / 1 failed** |
| `ZOMBIE_KILLER` | `(input, …)` | 11 ok |
| `ENROLLMENT_ORDER_GENERATOR` | `(input, …)` | 308 ok |

Guest garbage collection had therefore never once run since it was introduced.

## Fix

`addWork` has always normalised this with `input || {}`; `ensureOneWork` now does the same. That
asymmetry between the two work-creation paths was the actual defect. Because `ensureOneWork`
`$set`s onto the existing NEW document on each tick, it also repairs queued items in place.

Work queued by already-released versions still holds a null input and does not all self-heal, so
`WorkerDirector.doWork` settles it once at the dispatch boundary rather than leaving every
adapter to guard itself.

The adapters themselves are deliberately untouched — their `= {}` default is correct JavaScript;
they were simply being handed something the contract should never have produced.

## Verification on master

- Integration **999 pass, 0 fail**; unit **589 pass, 0 fail**; build + ESLint clean
- Reverting only `packages/` makes 2 of the 3 new tests fail. The third ("keeps an input the
  schedule does provide") passes either way by design — it is the control that proves the other
  two are not vacuous.

## Difference from the v4.8.x commit

Only the test. `v4.8.x` registers a probe adapter via `WorkerDirector.registerAdapter` /
`unregisterAdapter`; on `master` adapters come from `pluginRegistry`, which has no
single-plugin removal (only `clear()`, far too blunt for a shared test process). The test now
stubs `getAdapterByType` and restores it in a `finally`, which works the same on both branches
and mutates no process-wide state.

### DIFF
--- a/packages/core-worker/src/module/configureWorkerModule.ts
+++ b/packages/core-worker/src/module/configureWorkerModule.ts
@@ -534,7 +534,10 @@ export const configureWorkerModule = async ({ db, options }: ModuleInput<WorkerS
           query,
           {
             $set: {
-              input,
+              // Autoscheduled work usually carries no input, and an undefined `$set` value is
+              // stored as null rather than skipped. Keep the same shape `addWork` writes, so
+              // an adapter destructuring its input never faces a null.
+              input: input || {},
               worker: null,
               updated: created,
               retries,
--- a/packages/core/src/directors/WorkerDirector.ts
+++ b/packages/core/src/directors/WorkerDirector.ts
@@ -94,7 +94,9 @@ export const WorkerDirector: IWorkerDirector = {
     }
 
     try {
-      const output = await adapter.doWork(input, unchainedAPI, workId);
+      // Work queued by older versions can still hold a null input, which no default parameter
+      // guards against, so settle it here rather than in every adapter.
+      const output = await adapter.doWork(input ?? {}, unchainedAPI, workId);
       return output;
     } catch (error) {
       // DO not use this as flow control. The adapter should catch expected errors and return status: FAILED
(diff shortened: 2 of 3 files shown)


pozylon:
Closing in favour of cherry-picking from `v4.8.x` directly. The work lives there as 0308e31bf + 91e77249b.

Worth carrying over when porting: master resolves adapters through `pluginRegistry`, which has no single-plugin removal, so the test stubs `getAdapterByType` and restores it rather than registering a probe adapter. That version is already what sits on `v4.8.x`, so a straight cherry-pick works.

